PR #3, opened 2026-04-08
fix: polish instructions set after first day run.

Hyperkid123:


florkbr:
@Hyperkid123 the signing on this commit doesn't seem to be your usual GPG. That's fine just an FYI

memory-server/src/tools/tasks.py
```diff
@@ -9,7 +9,7 @@
 from ..models import Task
 
 ACTIVE_STATUSES = ("in_progress", "pr_open", "pr_changes")
-MAX_ACTIVE = 5
+MAX_ACTIVE = 10
```
florkbr, line 12: @Hyperkid123 assuming this is just expanding as we take on more tasks correct? Also of note, the CLAUDE.md mentions investigation tasks do not count towards this limit. Do we need a separate limit of investigation tasks as well?

memory-server/src/api.py
```diff
@@ -38,16 +48,35 @@ async def api_tasks(request: Request) -> JSONResponse:
...
+    if not row:
         return JSONResponse({"error": f"Task {jira_key} not found"}, status_code=404)
-    await bus.publish(Event("task_removed", {"jira_key": jira_key}))
-    return JSONResponse({"deleted": True, "jira_key": jira_key})
+    await bus.publish(Event("task_archived", {"jira_key": jira_key}))
+    return JSONResponse({"archived": True, "jira_key": jira_key, "task": _task(row)})
+
+
+async def api_task_unarchive(request: Request) -> JSONResponse:
+    """Restore an archived task back to paused so the bot can pick it up."""
```
florkbr, line 67: @Hyperkid123 assuming the docstring is just out of date here, but I think the code actually does the opposite — the docstring says "Restore an archived task back to paused" but the SQL sets status = 'in_progress'::task_status, paused_reason = NULL. **So it restores to in_progress and clears the paused reason.**

memory-server/src/tools/tasks.py
Hyperkid123: yeah just because we essentially threw hundreds of tickets at the bot. At some point, we can bring it back down. And also configure it via env.

PR #46, opened 2026-04-20
feat(infra): add HCC dev proxy for UI verification

Hyperkid123:
## Summary

- Add Caddy-based dev proxy for local UI verification against stage
- Build custom Caddy from source using `ubi9/go-toolset` multi-stage (passes Konflux Enterprise Contract — no docker.io base images)
- Plugins: `cache-handler`, `transform-encoder`, `rh_identity_transform` (JWT → x-rh-identity header)
- Update frontend persona: replace `fec dev` with `npm run build` → `npx http-server` → Caddy proxy → Chrome screenshots
- Allow Cypress CDN domains through Squid proxy

## How it works

Same pattern as [caddy-ubi](https://github.com/RedHatInsights/caddy-ubi): custom `main.go` imports plugins directly, built with `go build` on UBI go-toolset. No xcaddy, no docker.io images.

```dockerfile
FROM registry.access.redhat.com/ubi9/go-toolset:latest AS dev-proxy-builder
COPY dev-proxy/ /tmp/dev-proxy/
RUN cd /tmp/dev-proxy && go build -o /tmp/caddy .
```

## Test plan

- [x] Docker build succeeds — no EC-rejected base images
- [x] Caddy binary present at `/usr/local/bin/caddy` with correct version (v2.11.2)
- [x] All 3 plugins loaded (`caddy list-modules` shows cache, transform, rh_identity_transform)
- [x] Caddyfile validates (`caddy validate`)
- [x] Proxy serves Hybrid Cloud Console HTML from `console.stage.redhat.com` (200, 1703 bytes)
- [x] API routes proxy correctly (401 = stage auth rejection, expected without SSO cookie)
- [ ] Konflux EC pipeline passes

RHCLOUD-47012


Hyperkid123:
### Build from source (Konflux EC workaround)

The previous approach used `FROM quay.io/redhat-user-workloads/hcc-platex-services-tenant/frontend-development-proxy:latest` as a multi-stage base to copy the pre-built Caddy binary. However, Konflux Enterprise Contract rejects `quay.io/redhat-user-workloads` as a base image registry — even though that's where Konflux itself publishes images:

> Base image "quay.io/redhat-user-workloads/hcc-platex-services-tenant/frontend-development-proxy@sha256:..." is from a disallowed registry

Updated to build Caddy from source using `caddy:2.11.2-builder` + `xcaddy` with the same 3 plugins (cache-handler, transform-encoder, rh_identity_transform). The plugin source is vendored from [frontend-development-proxy](https://github.com/RedHatInsights/frontend-development-proxy).

Also added `download.cypress.io` and `cdn.cypress.io` to the Squid proxy allowlist for e2e test runner binary downloads.

Both changes verified: Docker build succeeds and all 8 Playwright e2e tests pass in the container.

dev-proxy/rh_identity_transform/main.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (rht *RhIdentityTransform) Validate() error {
+	return nil
+}
+
+func (rht RhIdentityTransform) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+	tokenStr, ok := ExtractToken(r)
+	if !ok {
+		next.ServeHTTP(w, r)
```
florkbr, line 40: Nit: missing `return` here — without it, execution falls through and `next.ServeHTTP` gets called a second time at line 49. Should be `return next.ServeHTTP(w, r)`.

florkbr: Actually, this is a bug in https://github.com/RedHatInsights/frontend-development-proxy too.

florkbr (approved):
@Hyperkid123 since we are trying to run fast this looks good, however I opened up a tech-debt ticket tracking all the duplicated code from https://github.com/RedHatInsights/frontend-development-proxy. See https://redhat.atlassian.net/browse/RHCLOUD-47145

PR #70, opened 2026-04-23
fix(bot): detect idle cycles for longer sleep interval

Hyperkid123:
## Summary

- Fix no-work detection that controls active (300s) vs idle (3600s) sleep interval
- Previously required exact string `NO_WORK_FOUND` in agent output, but the LLM outputs natural language like "nothing to pick up" or "no tickets"
- Now matches common no-work patterns in the agent's response

Without this fix, the bot sleeps 300s between every cycle even when there's no work, burning ~$0.30/cycle unnecessarily.

## Test plan
- [ ] Bot logs `Sleeping for 3600s` after cycles with no work
- [ ] Bot logs `Sleeping for 300s` after cycles that did work

### DIFF
--- a/bot/costs.py
+++ b/bot/costs.py
@@ -19,6 +19,24 @@ logger = logging.getLogger(__name__)
 COSTS_API = os.environ.get("COSTS_API_URL", "http://localhost:8080/api/costs")
 
 
+_NO_WORK_PATTERNS = [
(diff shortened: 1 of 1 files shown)


florkbr (approved):
Works as a stop gap. We will move to something more predictable ASAP

PR #116, opened 2026-04-29
fix(probes): follow up fix for proxy probe - use TCP

florkbr:
## Summary                                                                                                                                                                                                                                                                              
  - Switch proxy health check from HTTP (`curl`) to TCP (`/dev/tcp`) in the entrypoint
  - Squid doesn't serve HTTP on its proxy port — the previous `curl` check received `ERR_ACCESS_DENIED` even with `--noproxy '*'`, causing the bot to fail startup after 60s                                                                                                              
  - TCP check aligns with how the proxy is already probed in the OpenShift template (`tcpSocket` on port 3128)                                                                                                                                                                            
                                                                                                                                                                                                                                                                                          
  ## Test plan                                                                                                                                                                                                                                                                            
  - [x] Verified `bash -c "echo > /dev/tcp/devbot-proxy/3128"` succeeds from the bot pod                                                                                                                                                                                                  
  - [x] Verified `curl -sf --noproxy '*' http://devbot-memory-server:8080/health` still works for memory server                                                                                                                                                                           
  - [x] Local docker-compose unaffected — `PROXY_HOST` is not set, so the check is skipped (compose uses `depends_on: service_healthy`)

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -165,7 +165,7 @@ echo "Verifying GitLab auth..."
 glab auth status --hostname gitlab.cee.redhat.com 2>&1 | head -3 || { echo "WARNING: glab auth failed"; }
 
 # --- Wait for dependent services (OpenShift deploys all pods concurrently) ---
-wait_for() {
+wait_for_http() {
     local name="$1" url="$2" timeout="${3:-120}"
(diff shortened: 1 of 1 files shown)


florkbr:
Merging and will validate on the AI cluster (preventing bot startup)

PR #143, opened 2026-05-07
feat(proxy): Vertex AI auth proxy in executor-server

Hyperkid123:
## Summary

Move GCP service account key from bot container to proxy container. The executor-server now embeds an HTTP reverse proxy on port 8443 that injects OAuth2 Bearer tokens into Vertex AI requests.

- **New files**: `vertex.go` (reverse proxy), `vertex_policy.go` (model allowlist), plus tests
- **Bot sends unauthenticated requests** to `http://proxy:8443` with `CLAUDE_CODE_SKIP_VERTEX_AUTH=true`
- **Proxy rewrites** dummy project/region values, injects auth, forwards to Vertex AI
- **Model allowlist** configured via `VERTEX_ALLOWED_MODELS` env var (no hardcoded defaults)
- **Global endpoint** support: `aiplatform.googleapis.com` (not `global-aiplatform.googleapis.com`)
- **Bot cleanup**: removed SA key decode from entrypoint, GAC resolution from run.py, secret from config.py

### Requires

App-interface MR must merge first to set `GCP_PROJECT_ID`, `GCP_REGION`, `VERTEX_ALLOWED_MODELS` parameters:
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/186971

RHCLOUD-47293

## Test plan

- [x] `go test -v ./...` — all 17 tests pass
- [x] `docker compose build` — both images build
- [x] Proxy logs: `vertex-auth-proxy listening on :8443`
- [x] Healthz endpoint returns 200
- [x] rawPredict request returns status=200, Claude responds
- [x] Blocked model returns 403
- [x] Malformed path returns 400
- [x] Bot container has no SA key, no GCP credentials
- [x] Streaming responses work (FlushInterval: -1)

Generated with [Claude Code](https://claude.com/claude-code)


Dockerfile
```diff
@@ -141,7 +141,7 @@ RUN uv sync --frozen --no-dev
 ENV PATH="/home/botuser/app/.venv/bin:/home/botuser/go/bin:$PATH"
 ENV GOPATH="/home/botuser/go"
 ENV CLAUDE_CODE_USE_VERTEX=1
-ENV VERTEX_LOCATION=global
+ENV CLOUD_ML_REGION=global
```
florkbr, line 144: FWIW this may break local dev setups, but I've heard a few folks are already having issues with the new credentials harness pulling their global creds - so lets focus on getting everything running/complete on the deployed bot.

Hyperkid123: Yeah I think the days of local runs are over. You can always re-define this in your .env file. You wont be able to do this without the SA anymore. Cause thats what the proxy needs.

If we can share the SA with folks. But I think we are very close to the deployed runs. Or we can have a follow to make the local runs easier. But I don't want that to be the focus right this second.

proxy/executor/vertex.go
```diff
@@ -0,0 +1,111 @@
...
+				path = "/v1" + path
+			}
+			r.Out.URL.Path = rewritePath(path, projectID, region)
+			r.Out.URL.RawQuery = r.In.URL.RawQuery
+			r.Out.Host = upstream.Host
+
+			tok, err := ts.Token()
+			if err != nil {
+				log.Printf("vertex: token error: %v", err)
+				return
```
florkbr, line 49: If `ts.Token()` fails, the Rewrite func logs the error and returns, but I believe the request still gets proxied to Gcloud without an Authorization header. Gcloud will return a 401, but from the bot's perspective it's an opaque upstream error. We should probably also abort the request with a clear error instead of proxying it unauthenticated.

This is tricky in `httputil.ReverseProxy.Rewrite` since you can't return an error — you'd need to set a flag and check it in a ModifyResponse or ErrorHandler, or use a wrapping handler that validates the token before calling proxy.ServeHTTP.

Hyperkid123: I think we may handle this via monitoring/alerting later. If `ts.Token` fails the bot should be the least of our worries. It will not spend any tokens and the cycle should crash.

proxy/executor/vertex_test.go
```diff
@@ -0,0 +1,107 @@
...
+	req := httptest.NewRequest("POST", "/v1/bad/path", strings.NewReader("{}"))
+	w := httptest.NewRecorder()
+	handler.ServeHTTP(w, req)
+
+	if w.Code != http.StatusBadRequest {
+		t.Errorf("malformed path status = %d, want 400", w.Code)
+	}
+}
+
+func TestAllowedModelForwarded(t *testing.T) {
```
florkbr, line 89: @Hyperkid123 this seems like a dead test. It creates a TLS test server, assigns it to upstream, then immediately does `_ = upstream` and the function ends. The test never sends a request through the proxy? It never asserts anything.

florkbr (approved):

PR #154, opened 2026-05-11
feat(skills): add post-pr skill for efficient PR bookkeeping

catastrophe-brandon:
## Summary

Adds the `post-pr` skill to consolidate post-PR-creation bookkeeping into a single efficient operation, reducing 5-6 sequential tool calls into one script execution.

**JIRA**: [RHCLOUD-47263](https://issues.redhat.com/browse/RHCLOUD-47263)

## What This Skill Does

After creating a PR, the skill executes 6 operations in sequence:

1. **Update GitHub PR** - Adds labels (`code-review`, `awaiting-review`) and JIRA link
2. **Transition JIRA** - Moves ticket to "Code Review" status
3. **Add JIRA comment** - Posts PR link and summary
4. **Send Slack notification** - Notifies team of PR creation
5. **Store learnings** - Saves implementation notes to memory
6. **Update bot status** - Sets status to `idle`

## Usage

```bash
python3 .claude/skills/post-pr/post_pr.py <JIRA_KEY> [--dry-run]
```

Follows the same simple pattern as other skills in this repo (wrap-up, triage, new-work).

## Implementation Details

- **Wrapper script** (`post_pr.py`) - Fetches task from memory server, validates data, calls operations
- **Operations module** (`scripts/post_pr_operations.py`) - Core API integration logic (GitHub, JIRA, Slack)
- **Comprehensive error handling** - Graceful fallbacks for malformed inputs (URL normalization, type validation)
- **Professional logging** - Uses Python logging module, no print statements or emotes

## Testing

**62 tests total** (all passing):
- 35 operations/integration tests (ported from original implementation)
- 27 new wrapper tests (URL normalization, task validation, error handling)

### Local Testing
```bash
cd .claude/skills/post-pr
uv run pytest -v
```

### CI Pipeline
Tests run during the **Docker build phase** as part of the trusted artifact creation:
- Added to `Dockerfile` after `.claude/` directory is copied
- Executes `uv sync --frozen --all-extras && uv run pytest -v --tb=short`
- Build fails immediately if any tests fail
- Satisfies Enterprise Contract `trusted_task` policy (no untrusted inline tasks)

This approach ensures:
- ✅ Tests validate skill before image is finalized
- ✅ Passes Red Hat Enterprise Contract policies
- ✅ Fast failure on test errors (before security scans)
- ✅ Full test coverage integrated into trusted build

## Configuration

Requires environment variables:
- `GITHUB_TOKEN` or `GH_TOKEN`
[...]

PR #161, opened 2026-05-12
refactor(post-pr): replace API tokens with CLI tools and MCP

Hyperkid123:
## Summary

Addresses review feedback from PR #154 — the post-pr skill was still trying to use direct API tokens (Jira, GitHub, GitLab) that are no longer available in the container.

- **Jira**: Replaced `httpx` token auth with shared `jira_call()` MCP client
- **GitHub**: Replaced `httpx` + `GITHUB_TOKEN` with `gh api` CLI (pre-authenticated)
- **GitLab**: Added new `glab api --hostname` support (didn't exist before)
- **Slack**: Kept `httpx` — webhook URL is self-contained, no separate auth needed

Removes `github_token` parameter from constructor, workflow function, and CLI args. All 69 tests pass (up from 35 — added URL parsing, GitLab MR, and CLI failure tests).

RHCLOUD-47706

## Test plan

- [x] `uv run pytest -v` — all 69 tests pass
- [x] Container tested end-to-end by Martin
- [x] Verify `gh api` calls work in container
- [x] Verify `glab api` calls work for GitLab repos
- [x] Verify Slack webhook still fires

### DIFF
--- a/.claude/skills/post-pr/README.md
+++ b/.claude/skills/post-pr/README.md
@@ -80,7 +80,6 @@ result = execute_post_pr_workflow(
     ticket_id="RHCLOUD-456",
     summary="Add vector search caching",
     github_token=None,  # Falls back to GITHUB_TOKEN env var
-    jira_token=None,    # Falls back to POST_PR_JIRA_TOKEN env var
     slack_channel="#hcc-ai-assistant",
     reviewers=["user1", "user2"],
     skip_operations=[],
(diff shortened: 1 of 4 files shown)


karelhala (commented):
## Code Review

### Critical

**1. README example still references removed `github_token` parameter**

The README code example at line ~82 still includes `github_token=None` but `execute_post_pr_workflow()` no longer accepts that parameter. Anyone copying the example will get a `TypeError`.

**2. `_run_cli` conflates stdout and stderr**

```python
output = r.stdout.strip() or r.stderr.strip()
```

On success (returncode 0), if stdout is empty but stderr has warnings, stderr is returned as "output" — callers that `json.loads()` it will crash. On failure, if both stdout and stderr have content, stderr is silently lost (only stdout returned). Suggestion:

```python
output = r.stdout.strip()
error = r.stderr.strip()
if r.returncode != 0:
    return False, error or output  # prefer stderr for errors
return True, output
```

**3. GitHub URL parsing doesn't validate `/pull/` segment**

In `_parse_pr_url`, the GitHub branch only checks `len(parts) < 5` after splitting on `/`. A URL like `https://github.com/org/repo/issues/123` passes the length check and returns `owner="org", repo="repo"` without ever verifying that `parts[-2] == "pull"`.

**4. `jira_call` returns `None` for both "MCP not configured" and "runtime failure"**

In `jira_transition_issue`, the check `if not data` raises `"Failed to get transitions from Jira MCP"` but the actual cause could be that `JIRA_MCP_URL` is simply unset. Callers can't distinguish config errors from runtime failures. Consider checking `JIRA_MCP_URL` upfront or differentiating the error messages.

---

### Medium

**8. Hardcoded GitLab hostnames**

```python
if hostname == "gitlab.com" or hostname == "gitlab.cee.redhat.com":
```

Any other self-hosted GitLab instance is rejected as "Unsupported". Could detect GitLab by the `/-/merge_requests/` path pattern instead, which is GitLab-specific regardless of hostname.

**13. Hardcoded 30s timeout in `_run_cli`**

`subprocess.run` uses a hardcoded `timeout=30`. Some operations on slow GitLab instances (especially reviewer lookups that do user search + MR update) may need more time. Consider making this a class-level constant or a parameter with a default.

Hyperkid123:
@karelhala not doing the hard coded stuff now. It was a pre-existing issue. This focuses only on the jira/git access.

karelhala (approved):

PR #165, opened 2026-05-12
feat(config): remote config sync mechanism

Hyperkid123:
## RHCLOUD-47728

Adds remote config repo sync so bot instances can load personas, repos, skills, and MCP servers from an external git repo without rebuilding the container.

### What's new

**`rehor-config/agent/`** — initial HCC config directory:
- `personas/` — all current personas (frontend, backend, config, cve, operator, rbac, rds-upgrade, tooling)
- `project-repos.json` — full repo registry
- `mcp.json` — bot-level MCP server config
- `skills/` — empty dir for future custom skills

**`bot/run.py`** — sync mechanism:
- `sync_config_repo()` — clones (`--depth 1`) or pulls (`--ff-only`) the repo specified by `BOT_CONFIG_REPO` env var into `data/remote-config/`. `BOT_CONFIG_PATH` (default: `rehor-config`) specifies the subdirectory within the repo.
- `apply_remote_config()` — copies personas, project-repos.json into working dir. Skills and MCP servers are **additive only** — built-in core skills (triage, wrap-up, post-pr, new-work) and built-in MCP servers (mcp-atlassian) are never overwritten.
- Called at the start of each cycle, before `run_cycle()`
- Graceful failure: timeout/unreachable repo → warning logged, bot continues with built-in config

### Security boundaries (never overridden by remote config)
- `CLAUDE.md` — core workflow rules
- `.claude/settings.json` — hooks, permissions, sandbox
- `config.json` — model/interval settings
- Built-in skills and MCP servers

### Env vars
- `BOT_CONFIG_REPO` — git URL of config repo (optional)
- `BOT_CONFIG_PATH` — subdirectory within repo (default: `rehor-config`)

### DIFF
--- a/bot/run.py
+++ b/bot/run.py
@@ -104,6 +104,88 @@ def setup_logging() -> None:
     )
 
 
+REMOTE_CONFIG_DIR = DATA_DIR / "remote-config"
+
+
+def sync_config_repo(script_dir: Path) -> Path | None:
+    """Clone or pull BOT_CONFIG_REPO. Returns agent config dir or None."""
+    logger = logging.getLogger(__name__)
+    repo_url = os.environ.get("BOT_CONFIG_REPO")
+    if not repo_url:
+        return None
+
+    config_dir = REMOTE_CONFIG_DIR
+    try:
+        if (config_dir / ".git").exists():
+            subprocess.run(
+                ["git", "-C", str(config_dir), "pull", "--ff-only"],
+                capture_output=True, timeout=30, check=False,
+            )
+        else:
+            config_dir.parent.mkdir(parents=True, exist_ok=True)
+            subprocess.run(
+                ["git", "clone", "--depth", "1", repo_url, str(config_dir)],
+                capture_output=True, timeout=60, check=False,
+            )
+    except subprocess.TimeoutExpired:
+        logger.warning("Config repo sync timed out — using built-in config")
+        return None
+    except Exception as exc:
+        logger.warning("Config repo sync failed: %s — using built-in config", exc)
+        return None
+
+    sub = os.environ.get("BOT_CONFIG_PATH", "rehor-config")
+    agent_dir = config_dir / sub / "agent"
+    if not agent_dir.is_dir():
+        logger.warning("Config repo has no %s/agent/ dir — using built-in config", sub)
+        return None
+
+    return agent_dir
+
+
+def apply_remote_config(script_dir: Path, agent_dir: Path) -> None:
+    """Overlay remote agent config onto working directory."""
+    logger = logging.getLogger(__name__)
+
+    remote_personas = agent_dir / "personas"
+    if remote_personas.is_dir():
+        shutil.copytree(remote_personas, script_dir / "personas", dirs_exist_ok=True)
+        logger.info("Applied remote personas")
+
+    remote_repos = agent_dir / "project-repos.json"
+    if remote_repos.is_file():
+        shutil.copy2(remote_repos, script_dir / "project-repos.json")
+        logger.info("Applied remote project-repos.json")
+
+    remote_skills = agent_dir / "skills"
+    if remote_skills.is_dir():
+        shutil.copytree(
+            remote_skills, script_dir / ".claude" / "skills",
+            dirs_exist_ok=True,
+            ignore=shutil.ignore_patterns(".gitkeep"),
+        )
+        logger.info("Applied remote custom skills")
+
+    remote_mcp = agent_dir / "mcp.json"
+    if remote_mcp.is_file():
+        import json
+        bot_mcp_path = script_dir / "bot" / "mcp.json"
+        try:
+            with open(bot_mcp_path) as f:
+                built_in = json.load(f)
+            with open(remote_mcp) as f:
+                custom = json.load(f)
+            for name, cfg in custom.get("mcpServers", {}).items():
+                if name not in built_in.get("mcpServers", {}):
+                    built_in.setdefault("mcpServers", {})[name] = cfg
+            with open(bot_mcp_path, "w") as f:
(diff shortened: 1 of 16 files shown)


bot/run.py
```diff
@@ -104,6 +104,88 @@ def setup_logging() -> None:
...
+    if not repo_url:
+        return None
+
+    config_dir = REMOTE_CONFIG_DIR
+    try:
+        if (config_dir / ".git").exists():
+            subprocess.run(
+                ["git", "-C", str(config_dir), "pull", "--ff-only"],
+                capture_output=True, timeout=30, check=False,
+            )
```
karelhala, line 124: Both `git pull` (line 120) and `git clone` (line 127) use `check=False` with no inspection of `returncode` — if the operation fails (auth error, bad URL, disk full), it silently falls through. For a pull that fails mid-way, the bot could end up with corrupted/stale config and no warning.

Suggestion — check `returncode` and log `stderr` when non-zero:
```python
r = subprocess.run(
    ["git", "-C", str(config_dir), "pull", "--ff-only"],
    capture_output=True, timeout=30, check=False,
)
if r.returncode != 0:
    logger.warning("git pull failed (rc=%d): %s", r.returncode, r.stderr.decode().strip())
```
Same for the clone path below.

bot/run.py
```diff
@@ -104,6 +104,88 @@ def setup_logging() -> None:
...
+    return agent_dir
+
+
+def apply_remote_config(script_dir: Path, agent_dir: Path) -> None:
+    """Overlay remote agent config onto working directory."""
+    logger = logging.getLogger(__name__)
+
+    remote_personas = agent_dir / "personas"
+    if remote_personas.is_dir():
+        shutil.copytree(remote_personas, script_dir / "personas", dirs_exist_ok=True)
```
karelhala, line 164: `apply_remote_config()` reads and writes `bot/mcp.json` **in the source tree on disk** every cycle. This has a few consequences:

1. Custom MCP servers accumulate — if the remote config repo later removes one, it persists locally until container restart
2. In dev mode (git checkout), this dirties the working tree
3. Every cycle re-reads and re-writes the file even when nothing changed

Consider writing the merged config to `data/merged-mcp.json` (or a temp location) instead, and having the caller reference that. Or at least diff before writing and skip if unchanged.

bot/run.py
```diff
@@ -104,6 +104,88 @@ def setup_logging() -> None:
...
+
+    return agent_dir
+
+
+def apply_remote_config(script_dir: Path, agent_dir: Path) -> None:
+    """Overlay remote agent config onto working directory."""
+    logger = logging.getLogger(__name__)
+
+    remote_personas = agent_dir / "personas"
+    if remote_personas.is_dir():
```
karelhala, line 163: Nit: `import json` is already a top-level import in this file (used in `main()` for jira creds around line 307). This in-function import is unnecessary — just remove it.

rehor-config/agent/mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "mcp-atlassian": {
+      "type": "http",
+      "url": "${JIRA_MCP_URL}"
```
karelhala, line 5: `"${JIRA_MCP_URL}"` is a literal string here — it won't be shell-expanded when `apply_remote_config` merges this into `bot/mcp.json`. It's harmless today because the additive merge skips existing keys (so the built-in `mcp-atlassian` entry wins), but it's a misleading footgun: if someone removes `mcp-atlassian` from the built-in config, this broken literal would be injected.

Either add env var interpolation in `apply_remote_config`, or replace the placeholder with a comment explaining this file is a template that needs the real URL.

bot/run.py
```diff
@@ -104,6 +104,88 @@ def setup_logging() -> None:
     )
 
 
+REMOTE_CONFIG_DIR = DATA_DIR / "remote-config"
+
+
```
karelhala, line 110: Nit: `sync_config_repo` takes `script_dir` as a parameter but never uses it — the clone target is `REMOTE_CONFIG_DIR` (a module-level constant derived from `DATA_DIR`). Clean up the unused parameter, or use `script_dir` to derive the config dir if that was the intent.

karelhala (commented):
Nice work on the architecture — security boundaries are well-defined, graceful degradation is solid, and the additive merge for skills/MCP is the right call. A few things to address before merging:

rehor-config/agent/mcp.json
Hyperkid123: We expand the env var in other places :)

bot/run.py
Hyperkid123: Fixed — now checking returncode and logging stderr for both pull and clone paths. Clone failure returns None immediately; pull failure logs but continues (stale config is better than no config).

bot/run.py
Hyperkid123: Good catch — merged config now writes to `data/merged-mcp.json` instead of mutating the source tree. Also skips the write if content hasn't changed. `load_mcp_servers` in config.py picks it up (additive only, built-in keys still win).

bot/run.py
Hyperkid123: Moved to top-level import.

bot/run.py
Hyperkid123: Removed — `sync_config_repo()` now takes no parameters, uses module-level `REMOTE_CONFIG_DIR` directly.

rehor-config/agent/mcp.json
Hyperkid123: Acknowledged — skipping env var expansion for now as discussed. The additive merge logic means this entry won't be injected anyway (built-in `mcp-atlassian` key wins), but we'll address the placeholder in a follow-up.

karelhala (approved):

PR #178, opened 2026-05-14
feat(sleep): deterministic skill-based sleep signaling

Hyperkid123:
## Summary

- Skills now write `data/cycle-sleep.json` with recommended sleep duration instead of relying on fragile pattern matching of agent output text
- **Triage**: writes `recommended_sleep: 3600` when at capacity + all tasks clean
- **New-work**: writes `recommended_sleep: 3600` when no eligible candidates found
- No signal file = standard 300s sleep (work was done)
- Python runner reads and deletes the file after each cycle

RHCLOUD-47757

### DIFF
--- a/.claude/skills/new-work/new_work.py
+++ b/.claude/skills/new-work/new_work.py
@@ -13,6 +13,8 @@ import urllib.parse
 import urllib.request
 from pathlib import Path
 
+SLEEP_FILE = Path(__file__).resolve().parent.parent.parent.parent / "data" / "cycle-sleep.json"
+
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
 from jira_mcp import jira_call
 
@@ -186,6 +188,8 @@ def main():
     candidates = get_candidates()
     if not candidates:
         print("NO CANDIDATES FOUND")
+        SLEEP_FILE.parent.mkdir(parents=True, exist_ok=True)
+        SLEEP_FILE.write_text(json.dumps({"recommended_sleep": 3600, "reason": "no_eligible_work"}))
         return
 
     print(f"NEW WORK CANDIDATES ({len(candidates)})")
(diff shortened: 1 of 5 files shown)


.claude/skills/new-work/new_work.py
```diff
@@ -13,6 +13,8 @@
 import urllib.request
 from pathlib import Path
 
+SLEEP_FILE = Path(__file__).resolve().parent.parent.parent.parent / "data" / "cycle-sleep.json"
```
catastrophe-brandon: If there's a common path where we're keeping the agent's control metadata, we may want to define a constant for that, then reference it in the various file paths. Not a blocker for this PR, just something for consideration.

e.g.

BASE_DATA_PATH=Path(__file__).resolve().parent.parent.parent.parent / "data"
SLEEP_FILE = BASE_DATA_PATH / "cycle-sleep.json"

.claude/skills/triage/triage.py
```diff
@@ -13,6 +13,8 @@
 import urllib.request
 from pathlib import Path
 
+SLEEP_FILE = Path(__file__).resolve().parent.parent.parent.parent / "data" / "cycle-sleep.json"
```
catastrophe-brandon: See previous comment.

bot/run.py
```diff
@@ -156,13 +157,43 @@ def sync_config_repo() -> Path | None:
 
 
 
+SLEEP_SIGNAL_FILE = DATA_DIR / "cycle-sleep.json"
```
catastrophe-brandon, line 160: Funny; it was set up the desired way here, but not the other source files. Classic LLM inconsistency.

catastrophe-brandon (approved):
Minor comments, overall looks okay.

Hyperkid123:
@catastrophe-brandon thanks, it shuld be fixed now ;)

PR #195, opened 2026-05-20
fix(claim-ticket): inline constant to fix relative import

Hyperkid123:
## Summary
- Fixes relative import error in `claim_ticket_operations.py` that caused the bot to waste ~4 tool calls per cycle
- `from .constants import TRANSITION_IN_PROGRESS` fails when script is run directly with `python3` (not as package)
- Inlined the constant and deleted the now-unused `constants.py` module
- All 25 tests pass

RHCLOUD-47778


Hyperkid123:
Folded into #196

PR #196, opened 2026-05-20
fix: label-only JQL + claim-ticket import fix

Hyperkid123:
## Summary
- **new-work**: Drop `BOT_BOARD_NAME`/`BOT_BOARD_ID`/`BOT_SPRINT_PREFIX` and hardcoded `project = RHCLOUD` from candidate queries. Use label-only JQL with `sprint in openSprints()`. Three priority tiers: sprint/unassigned → sprint/bot-assigned → backlog.
- **claim-ticket**: Inline `TRANSITION_IN_PROGRESS` constant to fix `from .constants import` relative import error that wasted ~4 tool calls per cycle. Delete unused `constants.py`.

Board config (`BOT_BOARD_NAME`) remains used by claim-ticket for sprint assignment only.

RHCLOUD-47778

### DIFF
--- a/.claude/skills/claim-ticket/scripts/claim_ticket_operations.py
+++ b/.claude/skills/claim-ticket/scripts/claim_ticket_operations.py
@@ -29,7 +29,7 @@ import httpx
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent.parent))
 from jira_mcp import jira_call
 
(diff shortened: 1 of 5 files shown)


florkbr (approved):
Looks good; FWIW we should triple check we are setting the bot labels correctly as we are removing the RHCLOUD safeguard.

PR #213, opened 2026-05-26
feat: add auto-fork skill for automated repo forking

catastrophe-brandon:
Add new skill that automatically forks repos and updates config:
- Detects repos in project-repos.json needing forks
- Creates forks using gh repo fork (GitHub only)
- Updates config with fork URLs and commits changes
- Comprehensive test suite (23 tests, all passing)
- CI integration via GitHub Actions
- Docker build-time validation

Add pre-commit config with ruff for local linting/formatting.

Closes acceptance criteria:
- Bot detects repos without forks during triage
- Bot creates fork under its own GH account
- Bot opens PR to config repo with updated project-repos.json
- Works for both new repo additions and new instance configs
- GitLab repos handled (skipped with logged warning)

### Description

[RHCLOUD-47868](https://issues.redhat.com/browse/RHCLOUD-47868)

---

### Blast radius

---

### Rollback plan

---

### Checklist

### AI disclosure

[RHCLOUD-47868]: https://redhat.atlassian.net/browse/RHCLOUD-47868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/skills/auto-fork/SKILL.md
Hyperkid123, line 1: Can you compress this with caveman?

.claude/skills/auto-fork/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+# Bot's GitHub account (required) — username for fork destination
+# Must be a valid GitHub username (alphanumeric, hyphens, max 39 chars)
+export BOT_GITHUB_USERNAME=platex-rehor-bot
+```
+
+Optional environment variables:
+
+```bash
+# Instance ID (optional) — branch will be bot/auto-fork-{instance_id}
+export BOT_INSTANCE_ID=rehor
```
Hyperkid123: These could be extracted already from the remote config env variables. We can load them directly in the Python skill. The bot does not have to set these.

.claude/skills/auto-fork/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+# Config path (optional) — defaults to "rehor-config"
+export BOT_CONFIG_PATH=rehor-config
+```
+
+## Repo Detection
+
+Identifies repos needing forks:
+- Repos with `upstream` defined but `url` doesn't match bot's account pattern
+- For GitHub: checks if `url` contains `github.com/{BOT_GITHUB_USERNAME}/`
+- GitLab repos logged and skipped (no auto-fork support yet)
```
Hyperkid123: Can we instruct it to use the glab api cli?

.claude/skills/auto-fork/SKILL.md
catastrophe-brandon: Unclear what is meant here. The skills pull their configs from env vars defined in the proxy container. Are we trying to do away with those env vars totally? I think Claude was confused as well:

```
The docs say these are "auto-provided by bot runtime" but the reviewer's comment suggests they might not need to be env vars at
  all.

  Looking at the code:
  - BOT_CONFIG_PATH (line 88) — used to find data/remote-config/{config_path}/agent/project-repos.json
  - BOT_INSTANCE_ID (line 87) — used only for branch naming bot/auto-fork-{instance_id}

  The reviewer might be saying: instead of passing these as env vars, the skill could discover them:
  - BOT_CONFIG_PATH → scan data/remote-config/ for the subdirectory that exists
  - BOT_INSTANCE_ID → maybe extract from git config or not needed?

  Should I refactor auto-fork to auto-discover these values instead of reading env vars? Or is the current approach (using runtime-provided env vars from deploy/template.yaml)
  acceptable?

  What did the reviewer specifically mean by "extract from remote config"?
```

We may want to explicitly call out our methodology for discovering/configuring these values in the project-wide CLAUDE.md or some other file so that it's clear.

.claude/skills/auto-fork/SKILL.md
catastrophe-brandon: Updated.

.claude/skills/auto-fork/SKILL.md
catastrophe-brandon: Updated.

.claude/skills/auto-fork/SKILL.md
catastrophe-brandon: Added discovery in `b1d6484`, but I think it would be wise to revert that change.

Hyperkid123 (changes requested):
## Review

Good foundation — the detection + forking logic is solid, tests are comprehensive, and GitLab support is a nice addition. A few issues need addressing before merge:

### 1. No PR creation to config repo (blocking)

The skill commits locally to `data/remote-config/` but never pushes or opens a PR. The SKILL.md says "use push-and-pr skill" as step 3, but that's a manual agent step with no automation or enforcement. The whole point of this skill is to update the config repo — without the PR, the local commit is lost on next pod restart (the config repo gets re-cloned fresh each cycle via `sync_config_repo()` in `run.py`).

Options:
- Add push + PR creation to the script itself (preferred — self-contained)
- Or at minimum, have the SKILL.md `allowed-tools` include push-and-pr invocation and make step 3 an explicit instruction the agent executes

### 2. Ignores `BOT_CONFIG_PATH` and `BOT_CONFIG_REPO` env vars

`run.py` uses `BOT_CONFIG_REPO` (the git remote URL) and `BOT_CONFIG_PATH` (subdirectory name, default `rehor-config`) to locate the config. The auto-fork script ignores both and auto-discovers by iterating `data/remote-config/*/agent/`. This:

- Could find the wrong subdir if multiple configs exist
- Can't push because it doesn't know the remote URL came from `BOT_CONFIG_REPO`
- Contradicts the CLAUDE.md rule this very PR adds: "use runtime-provided env vars... `BOT_CONFIG_PATH` (already exists)"

Should use `BOT_CONFIG_PATH` directly (like `run.py` does) instead of iterating.

### 3. Instance ID re-derived instead of using `BOT_INSTANCE_ID`

The script extracts instance_id by splitting the config dir name on `-` and taking parts after index 2. But `BOT_INSTANCE_ID` is already an env var in the deployment template (line 432). The splitting logic is also brittle — `some-other-config-name` (4 parts) yields `config-name` instead of the actual instance ID.

### 4. `os.chdir()` mutates global process state

`_create_feature_branch()` calls `os.chdir(config_work_dir)` which changes the working directory for the entire process. If anything fails after this, the bot is in the wrong directory. Use `cwd=` parameter on subprocess calls instead.

### 5. Config file modified before branch creation

In `update_and_commit()`, `_update_config_file()` runs before `_create_feature_branch()`. This modifies `project-repos.json` on the default branch. If branch creation then fails, the working tree is dirty on master. Should create the branch first, then modify the file.

### 6. CI workflow env var mismatch

`.github/workflows/test-skills.yml:59` sets `BOT_GITHUB_USERNAME: test-bot` but the code was refactored to use `GH_USER_NAME`. The test fixtures set `GH_USER_NAME` via monkeypatch so unit tests pass, but the workflow env var is stale/unused.

---

Items 1-3 are the important ones — the rest are straightforward fixes.

.claude/skills/auto-fork/SKILL.md
catastrophe-brandon: Reverted discovery.

justinorringer (approved):
Looking good!

Hyperkid123 (commented):
### Review Notes

**1. Loose username match in `detect_unforkable_repos`** (`auto_fork.py`)
```python
if current_url and bot_user and bot_user in current_url:
```
This uses a substring check. If the bot username is short (e.g. `"a"`), it matches any URL containing that character. Should use a path-segment check like `f"/{bot_user}/"` or parse the URL and compare the path owner segment.

**2. `_detect_host_from_url` silently overrides explicit config** (`auto_fork.py`)
```python
host = config.get("host", HOST_GITHUB)  # reads from config
# ...
host = self._detect_host_from_url(upstream)  # immediately overwrites
```
The explicit `host` field from `project-repos.json` is read then discarded on the next line. Either remove the first read or use it as primary with URL detection as fallback.

**3. `_create_feature_branch` crashes on re-run** (`auto_fork.py`)
```python
subprocess.run(["git", "checkout", "-b", branch_name], ...)
```
If `bot/auto-fork` already exists from a previous run, `-b` fails with `CalledProcessError`. The `update_and_commit` catch block handles it, but the operation fails silently. Use `-B` instead, or delete the branch first.

**4. `git pull --ff-only` assumes tracking branch** (`auto_fork.py`)
```python
subprocess.run(["git", "pull", "--ff-only"], cwd=config_work_dir, check=True, ...)
```
If the config repo was freshly cloned and the local branch does not track an upstream, this fails. Since `git fetch origin` already ran, consider `git reset --hard origin/{default_branch}` instead — it is both simpler and does not require tracking to be set up.

**5. Inconsistent mock attributes in `test_workflow_partial_failure`** (`tests/test_integration.py`)
The `gh fork` branches in `mock_run_side_effect` set `returncode` and `stderr` but not `stdout`. The `else` branch sets all three. Mock auto-creates missing attributes as Mock objects so tests pass, but if the production code ever reads `stdout` from a fork result it would get a Mock instead of a string. Minor, but worth keeping consistent.

**6. README.md duplicates SKILL.md** 
101-line README + 57-line SKILL.md with substantial overlap. Other skills in the repo only have SKILL.md. Consider dropping the README or trimming it to dev-only content (install, test commands) that is not in SKILL.md.

Hyperkid123:
**7. Missing core use case: discovering and adding new repos**

The skill currently only processes repos that are already in `project-repos.json` but have the wrong `url`. The primary use case — bot encounters a `repo:` label on a Jira ticket for a repo not yet in its config, forks it, and adds it to `project-repos.json` — is not implemented.

The full flow should be:
1. Bot picks up a ticket with `repo:some-repo` label
2. `some-repo` is not in `project-repos.json` → detected as needing setup
3. Resolve the upstream URL from the `repo:` label (e.g. `repo:insights-chrome` → `https://github.com/RedHatInsights/insights-chrome`)
4. Check if bot already has a fork → if not, create one
5. Add new entry to `project-repos.json` with `upstream`, `url` (fork), and `host`
6. Commit, push, open PR to config repo

Currently only steps 4-6 work, and only for entries that already exist. The discovery and entry creation (steps 1-3, 5) are missing — which is the main reason this skill exists.

PR #234, opened 2026-06-10
feat(memory-server): add cycle_runs table for progress and transcripts

Hyperkid123:
## Summary

- Adds `cycle_runs` table to memory-server DB for storing per-cycle progress history (JSONB) and compressed transcripts (BYTEA, zstd)
- New REST API: `POST/GET /api/cycle-runs` for CRUD, `GET /api/cycle-runs/{id}/transcript` for on-demand decompression
- New MCP tools: `progress_store` and `progress_load` for bot agent to persist/retrieve structured progress across cycles
- 18 new tests covering all endpoints and MCP tools

## Context

[RHCLOUD-48343](https://redhat.atlassian.net/browse/RHCLOUD-48343)

The bot currently has no persistent record of what happened in each cycle beyond a single task state snapshot. This adds:
1. **Progress history** — structured JSONB summaries per cycle, loadable by the bot next cycle for continuity
2. **Transcript storage** — full zstd-compressed JSONL transcripts for human debugging and optimization

Both REST API and MCP tools accept `task_id` directly (DB integer PK) — no jira_key coupling. Null `task_id` for idle/error cycles.

## Test plan

- [x] `pytest tests/test_cycle_runs.py` — 18 tests passing
- [x] `pytest tests/` — all 26 tests passing (no regressions)
- [x] Live validation against running memory-server: POST, GET with filters, transcript round-trip (compress → store → decompress → verify match)

[RHCLOUD-48343]: https://redhat.atlassian.net/browse/RHCLOUD-48343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/memory-server/tests/test_cycle_runs.py
+++ b/memory-server/tests/test_cycle_runs.py
@@ -111,7 +111,9 @@ async def test_post_cycle_run_with_transcript(mock_pool):
 @pytest.mark.asyncio
 async def test_post_cycle_run_no_task(mock_pool):
     mock_pool.fetchrow = AsyncMock(
-        side_effect=lambda q, *a: _fake_cycle_run_row(id=2, task_id=None, cycle_type="idle")
+        side_effect=lambda q, *a: _fake_cycle_run_row(
+            id=2, task_id=None, cycle_type="idle"
+        )
     )
     body = {
         "cycle_type": "idle",
@@ -311,7 +313,9 @@ def mock_pool_for_tools():
     pool.fetch = AsyncMock(
         return_value=[
             _fake_cycle_run_row(id=1),
-            _fake_cycle_run_row(id=2, progress=json.dumps({"last_step": "tests_passing"})),
+            _fake_cycle_run_row(
+                id=2, progress=json.dumps({"last_step": "tests_passing"})
+            ),
         ]
     )
     with patch("src.tools.cycles.get_pool", return_value=pool):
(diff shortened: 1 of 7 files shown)


memory-server/tests/test_cycle_runs.py
```diff
@@ -0,0 +1,405 @@
...
+    async with AsyncClient(
+        transport=ASGITransport(app=app), base_url="http://test"
+    ) as client:
+        resp = await client.post("/api/cycle-runs", json=body)
+
+    assert resp.status_code == 201
+    data = resp.json()
+    assert data["id"] == 1
+    assert data["task_id"] == 42
+    assert data["cycle_type"] == "task_work"
```
charlesmulder, line 81: Might be better to reuse the `body` hash instead of repeating the values ie. `assert data["cycle_type"] == body["cycle_type"]`. Keeping it DRY.

charlesmulder (approved):

memory-server/tests/test_cycle_runs.py
Hyperkid123: I'd like to keep it as is because these are tests. Its easier to understand the assertions when the values are nice and static. Plus if we share variable, we may end up missing some regressions as we change variables and assertions where we did not intended.

PR #264, opened 2026-06-24
docs: preset system design doc

Hyperkid123:
## Summary

Design doc for the composable preset system (RHCLOUD-48671, epic RHCLOUD-48670).

Replaces the monolithic CLAUDE.md with layered presets:
- **Workflow presets** (1 per instance): the decision loop (jira-sprint, jira-kanban, etc.)
- **Env presets** (0–N per instance): additive capabilities (browser, container-scan, slack)
- **Shared skills**: reusable across workflows (push-and-pr, post-pr, auto-fork)
- **Multi-profile config repos**: same image, same config repo, different `BOT_CONFIG_PATH` per deployment
- **Build-time install scripts**: env presets include `install.sh` for system deps
- **Custom workflows**: `./` prefix in `instance.yaml` loads from config repo instead of built-in

Covers: directory layout, manifest format, CLAUDE.md decomposition, merge order, startup validation, migration path, open questions.

## Test plan

- [ ] Review design doc for completeness
- [ ] Validate against existing instance repos (framework, fleetshift, kessel)
- [ ] Resolve open questions before cutting implementation tickets

### DIFF
--- a/docs/presets-design.md
+++ b/docs/presets-design.md
@@ -1,0 +1,953 @@
+# Preset System Design
+
+Composable configuration system replacing the monolithic CLAUDE.md with layered presets that instances mix and match.
+
+**Epic**: RHCLOUD-48670  
+**Spike**: RHCLOUD-48671
+
+---
+
+## Problem
+
+Today every instance runs the same 432-line `CLAUDE.md` — the full Jira-triage-implement-PR workflow. But not every instance needs that. A "browser" instance that only does visual QA doesn't need the implementation loop. A future "reviewer" instance needs PR review logic but not Jira ticket claiming. The monolith forces all behavior onto every instance.
+
+Skills are partially decoupled (remote config can add custom skills), but the core workflow is a single file. Personas are per-tech-stack, but the decision engine that uses them is baked into the workflow.
+
+## Concepts
+
+### Preset Types
+
+| Type | Cardinality | Purpose |
+|------|-------------|---------|
+| **Workflow** | Exactly 1 per instance | The main CLAUDE.md — defines the bot's decision loop (triage → implement → PR maintenance, or review-only, or visual QA, etc.) |
+| **Env** | 0–N per instance | Additive capabilities: browser/chrome-devtools, container scanning, code review tools. Each adds MCP servers, skills, settings, or environment setup. |
+
+A workflow preset is the "brain" — what the bot does each cycle. Env presets are "hands" — tools and capabilities it can use.
+
+### Why Not Just Multiple CLAUDE.md Files?
+
+The workflow isn't just CLAUDE.md. It's CLAUDE.md + skills + personas + MCP servers + settings that all work together. A "review" workflow needs different skills than the "implement" workflow. Presets bundle everything a workflow or capability needs into one unit.
+
+## Directory Layout
+
+### Core Repo (`platform-frontend-ai-dev`)
+
+```
+presets/
+├── workflows/
+│   ├── jira-sprint/                   # Current workflow (Jira sprints → triage → implement → fork PR)
+│   │   ├── CLAUDE.md                  # Decision loop, priorities, Jira integration
+│   │   ├── skills/                    # Workflow-specific skills only
+│   │   │   ├── triage/
+│   │   │   ├── new-work/
+│   │   │   ├── claim-ticket/
+│   │   │   └── wrap-up/
+│   │   ├── preflight/                 # Pre-session scripts
+│   │   │   ├── 01-triage.py
+│   │   │   └── 02-find-work.py
+│   │   └── manifest.yaml
+│   │
+│   ├── reviewer/                      # Future: PR review only
+│   │   ├── CLAUDE.md
+│   │   ├── skills/
+│   │   └── manifest.yaml
+│   │
+│   └── investigator/                  # Future: Investigation/analysis only
+│       ├── CLAUDE.md
+│       ├── skills/
+│       └── manifest.yaml
+│
+├── shared/                            # Skills reusable across workflows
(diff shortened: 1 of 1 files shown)


Jakub007d (approved):

lenasolarova (approved):

docs/presets-design.md
```diff
@@ -0,0 +1,926 @@
...
+      BOT_LABEL: hcc-ai-fleetshift-community
+      BOT_INSTANCE_ID: 'Věnceslav Říční z Komunální lodi'
+      BOT_CONFIG_REPO: https://github.com/fleetshift/fleetshift-bot-instance.git
+      BOT_CONFIG_PATH: instance/community         # ← picks community profile
+      # ...
+```
+
+Same image, same config repo, two deployments. The only difference is `BOT_CONFIG_PATH` and `BOT_NAME`/`BOT_LABEL`/`BOT_INSTANCE_ID`. Each profile's `instance.yaml` selects its workflow, source, and envs. Both inherit shared personas and project-repos via `shared: ../shared`.
+
+**`shared/` convention**: `run.py` looks for a `shared/` directory as a sibling of the profile dir (i.e. `BOT_CONFIG_PATH/../shared/`). If it exists, shared assets are merged first, then profile-specific assets overlay on top. No config needed — no `shared/` dir means nothing to inherit.
```
florkbr: I think it's worth noting `shared/` sibling convention is implicit - this is clean for the happy path but invisible when debugging. If someone renames or moves the shared/ dir, the bot silently loses shared personas with no error.

docs/presets-design.md
```diff
@@ -0,0 +1,926 @@
...
+
+## Problem
+
+Today every instance runs the same 432-line `CLAUDE.md` — the full Jira-triage-implement-PR workflow. But not every instance needs that. A "browser" instance that only does visual QA doesn't need the implementation loop. A future "reviewer" instance needs PR review logic but not Jira ticket claiming. The monolith forces all behavior onto every instance.
+
+Skills are partially decoupled (remote config can add custom skills), but the core workflow is a single file. Personas are per-tech-stack, but the decision engine that uses them is baked into the workflow.
+
+## Concepts
+
+### Preset Types
```
florkbr, line 18: I think the preset concept is sound. I would just suggest we implement it after the CLAUDE.md decomposition (Phase 1) and env presets (Phase 2) first. This will make it so we only design the workflow preset abstraction when the second workflow is actually being built — otherwise we'll be guessing/speculating at the interface.

docs/presets-design.md
```diff
@@ -0,0 +1,926 @@
...
+| **Workflow** | Exactly 1 per instance | The main CLAUDE.md — defines the bot's decision loop (triage → implement → PR maintenance, or review-only, or visual QA, etc.) |
+| **Env** | 0–N per instance | Additive capabilities: browser/chrome-devtools, container scanning, code review tools. Each adds MCP servers, skills, settings, or environment setup. |
+
+A workflow preset is the "brain" — what the bot does each cycle. Env presets are "hands" — tools and capabilities it can use.
+
+### Why Not Just Multiple CLAUDE.md Files?
+
+The workflow isn't just CLAUDE.md. It's CLAUDE.md + skills + personas + MCP servers + settings that all work together. A "review" workflow needs different skills than the "implement" workflow. Presets bundle everything a workflow or capability needs into one unit.
+
+## Directory Layout
```
florkbr, line 31: In the same vein as my previous comment (more about implementation order than anything):

Shared_skills adds a third merge tier for 3 skills. It's low cost to include now, but also low cost to add later when a second workflow consumer actually exists. Either way is fine.

docs/presets-design.md
```diff
@@ -0,0 +1,926 @@
...
+Minimal instance (no browser, no slack):
+
+```yaml
+# fleetshift-config/internal/agent/instance.yaml
+workflow: jira-sprint
+source: jira
+envs: []
+```
+
+### Multi-Profile Config Repos
```
florkbr, line 347: No team currently runs two profiles from one repo. I'm curious if someone is trying to onboard with this use case in mind. If not but we do expect the need, I would just want to make sure this ships tested for when someone does decide to use it.

docs/presets-design.md
```diff
@@ -0,0 +1,926 @@
...
+2. load_instance_config()                # NEW: Read instance.yaml
+3. resolve_presets()                      # NEW: Find workflow + env presets
+4. assemble_claude_md()                   # NEW: Concatenate core + workflow + instance
+5. merge_preset_assets()                  # NEW: Merge skills, MCP, settings, hooks
+6. apply_merged_config()                  # Existing: merge remote config (personas, repos, etc.)
+7. validate_startup()                     # NEW: Fail if no CLAUDE.md assembled
+8. run_cycle()                            # Existing: start agent loop
+```
+
+### Validation Rules
```
florkbr, line 580: This is actually a good argument for the manifest system. We may want Phase 1 to include the validation infrastructure or we're just moving config around without gaining safety.

docs/presets-design.md
Hyperkid123: There are a few teams that want some triage/repeating task. I don't think we should require new konflux (runner image) onboarding.

docs/presets-design.md
Hyperkid123: That's why we have this design. We already have two presents. Jira sprint workflow and Jira kanban workflow. There are differences that make it not ideal to combine them. We were also asked about GH workflow but that is not yet required.

But agree with the order.

Hyperkid123:
@florkbr agreed with the incremental implementation. That was always the plan. I planned to reflect it in Jira but I will add a bit more detail to this doc.

PR #270, opened 2026-06-25
fix(memory-server): link transcript to existing cycle_run

Hyperkid123:
## Summary

- When `progress_store` creates a `cycle_runs` record mid-cycle and `record_transcript` uploads the transcript afterward, the API was creating a second orphan row instead of attaching to the existing one
- Now the API first tries to UPDATE the most recent cycle_run for the same instance that has no transcript (within 2h), falling back to INSERT if none found
- 72 existing orphan records across all instances

Fixes RHCLOUD-48739

## Test plan

- [x] All 20 cycle_runs tests pass (including new `test_post_cycle_run_transcript_links_to_orphan`)
- [x] Ruff lint + format clean
- [ ] Deploy and verify new cycles produce single merged records instead of orphan pairs

### DIFF
--- a/memory-server/bot_memory_server/api.py
+++ b/memory-server/bot_memory_server/api.py
@@ -903,23 +903,57 @@ async def api_cycle_runs_add(request: Request) -> JSONResponse:
     parsed_started = datetime.fromisoformat(started_at) if started_at else None
     parsed_finished = datetime.fromisoformat(finished_at) if finished_at else None
 
-    row = await pool.fetchrow(
-        f"""
-        INSERT INTO cycle_runs (task_id, cycle_type, instance_id, started_at, finished_at,
-                                tool_calls, tokens_used, progress, transcript)
-        VALUES ($1, $2, $3, COALESCE($4, NOW()), $5, $6, $7, $8, $9)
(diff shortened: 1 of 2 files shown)


memory-server/bot_memory_server/api.py
```diff
@@ -903,23 +903,57 @@ async def api_cycle_runs_add(request: Request) -> JSONResponse:
...
+            body.get("tool_calls"),
+            body.get("tokens_used"),
+            task_id,
+            instance_id_val,
+        )
+
+    if row is None:
+        row = await pool.fetchrow(
+            f"""
+            INSERT INTO cycle_runs (task_id, cycle_type, instance_id, started_at, finished_at,
```
florkbr, line 942: @Hyperkid123 Non-blocking future optimization: if `cycle_runs` grows large, the orphan-finding subselect could benefit from a partial index: `CREATE INDEX ON cycle_runs (instance_id, created_at DESC) WHERE transcript IS NULL;`

Not needed at current scale. Alternatively, a retention policy that prunes old cycle_runs (or drops transcript data after N days) would keep the table small enough that this never matters.

Food for thought. Not blocking so :shipit:

florkbr (approved):

PR #285, opened 2026-06-30
feat(preflight): add pre-session script runner

Hyperkid123:
## Summary

Adds preflight infrastructure that runs Python scripts **before** starting a Claude session, saving token costs on idle/error cycles (~$0.50-2.00/cycle when nothing to do).

**RHCLOUD-48706** (runner infra) + **RHCLOUD-48707** (preflight scripts)

### Architecture

- **`bot/preflight.py`** — runner module: discovers, executes, and aggregates preflight scripts
- **`presets/shared/preflight/`** — reusable modules (GH PR status, GL MR status, Jira triage, common utilities) importable by any workflow
- **`presets/workflows/jira-sprint/preflight/`** — 4-script pipeline: thin entry points importing shared modules + workflow-specific find-work
- **JSON protocol**: each script outputs `{"status": "start|skip|error", "content": "..."}`
- **Inter-script state**: `data/preflight-state.json` caches tasks/capacity across scripts

### Integration

- `bot/run.py` — preflight runs before session; skip/error bypass Claude entirely (0 tokens)
- `bot/agent.py` — `run_cycle()` accepts `preflight_prompt` to inject pre-gathered data
- `bot/transcripts.py` — `post_orphan_cycle()` gives dashboard visibility into skipped cycles

### Composability

Workflows pick which shared scripts to include. A GitHub-only workflow skips GL, a Jira-only skips both forge scripts. Entry points are 3-line re-exports:

```python
#!/usr/bin/env python3
from gh_pr_status import main
main()
```

### Tests

25 tests in `bot/tests/test_preflight.py` covering discovery, script execution (start/skip/error/invalid JSON/timeout/unknown status), aggregation logic, PYTHONPATH injection, and end-to-end `run_preflight()`.

## Test plan

- [x] `ruff check bot/ presets/` — clean
- [x] `python3 -m pytest bot/tests/ -v` — 55/55 pass (including 25 new)
- [ ] Konflux Tekton pipeline passes (existing `run-skill-tests` task runs pytest)
- [ ] Manual integration: full cycle with preflight → idle cycle skips Claude session

### DIFF
--- a/bot/preflight.py
+++ b/bot/preflight.py
@@ -1,0 +1,156 @@
+"""Pre-flight script runner — executes data-gathering scripts before Claude session.
+
+Each script prints JSON to stdout: {"status": "start"|"skip"|"error", "content": "..."}
+- start: work found, content becomes session prompt
+- skip: nothing to do, content logged as orphan cycle
+- error: script failed, content logged, backoff applied
+
+Aggregation: any error → no session. Any start → session starts. All skip → no session.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import subprocess
+from dataclasses import dataclass, field
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+PREFLIGHT_TIMEOUT = 120
+STATE_FILENAME = "preflight-state.json"
+
+
+@dataclass
+class ScriptResult:
+    name: str
+    status: str  # "start" | "skip" | "error"
+    content: str
+
+
+@dataclass
+class PreflightResult:
+    action: str  # "start" | "skip" | "error"
+    prompt: str = ""
+    transcript: str = ""
+    scripts: list[ScriptResult] = field(default_factory=list)
+
+
+def discover_preflight_scripts(
+    script_dir: Path,
+    workflow: str,
+    remote_agent_dir: Path | None = None,
+) -> list[Path]:
+    """Find preflight scripts from workflow preset + instance config, sorted by name."""
+    scripts: list[Path] = []
+
+    workflow_preflight = script_dir / "presets" / "workflows" / workflow / "preflight"
+    if workflow_preflight.is_dir():
+        scripts.extend(sorted(f for f in workflow_preflight.iterdir() if f.suffix == ".py" and f.is_file()))
+
+    if remote_agent_dir:
+        instance_preflight = remote_agent_dir / "preflight"
+        if instance_preflight.is_dir():
+            scripts.extend(sorted(f for f in instance_preflight.iterdir() if f.suffix == ".py" and f.is_file()))
+
+    return scripts
+
+
+def _run_script(script: Path, script_dir: Path) -> ScriptResult:
+    """Run a single preflight script and parse its JSON output."""
+    name = script.name
+    env = os.environ.copy()
+    shared_preflight = script_dir / "presets" / "shared" / "preflight"
+    skills_dir = script_dir / ".claude" / "skills"
+    extra = [str(shared_preflight), str(skills_dir)]
+    existing = env.get("PYTHONPATH", "")
+    env["PYTHONPATH"] = os.pathsep.join(extra + ([existing] if existing else []))
+    try:
+        proc = subprocess.run(
+            ["python3", str(script)],
+            capture_output=True,
+            text=True,
+            timeout=PREFLIGHT_TIMEOUT,
+            cwd=str(script_dir),
+            env=env,
+        )
+    except subprocess.TimeoutExpired:
+        logger.error("Preflight %s timed out after %ds", name, PREFLIGHT_TIMEOUT)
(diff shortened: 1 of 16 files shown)


presets/shared/preflight/common.py
```diff
@@ -0,0 +1,207 @@
...
+def fmt_comments(comments, label, since=None):
+    """Format a list of comments, optionally filtering by timestamp."""
+    if not comments:
+        return f"  {label}: (none)"
+    if since:
+        since_prefix = since[:16]
+        comments = [c for c in comments if (c.get("t", c.get("created", ""))[:16]) > since_prefix]
+    if not comments:
+        return f"  {label}: (none since last_addressed)"
+    comments = list(reversed(comments))
```
florkbr, line 135: We may want to consider a upper limit for comments. For example a ticket with 80 comments dumps all of them into the session prompt. The since filter helps when last_addressed is set, but for new tasks or tasks without last_addressed, everything comes through

bot/preflight.py
```diff
@@ -0,0 +1,156 @@
...
+
+def _aggregate(results: list[ScriptResult]) -> PreflightResult:
+    """Aggregate script results: any error → error, any start → start, all skip → skip."""
+    has_error = any(r.status == "error" for r in results)
+    has_start = any(r.status == "start" for r in results)
+
+    all_content = "\n\n".join(r.content for r in results if r.content)
+
+    if has_error:
+        return PreflightResult(action="error", transcript=all_content, scripts=results)
```
florkbr, line 131: I think this means a transient GitHub API failure (e.g. gh pr view times out) blocks the entire cycle, even if Jira triage found actionable feedback correct?

I think it makes sense for now, but could be softened later to "exclude errored scripts, aggregate the rest" — so a GH outage doesn't block Jira work. Fine for now, just might be worth documenting the tradeoff

presets/shared/preflight/gl_mr_status.py
```diff
@@ -0,0 +1,229 @@
...
+        elif "closed" in issues:
+            closed.append(e)
+        elif any(i.startswith("ci_fail") for i in issues):
+            ci_fail.append(e)
+        elif "conflict" in issues:
+            conflict.append(e)
+        elif "unresolved_threads" in issues:
+            feedback.append(e)
+        else:
+            clean.append(e)
```
florkbr, line 196: Are we missing `has_new_feedback` parity with GH? In gh_pr_status.py, there's a fallback before the clean bucket:

```
elif has_new_feedback(e):
    feedback.append(e)
```

This catches new non-bot PR comments since last_addressed even when classify_gh didn't flag any issues. The GL version here goes straight to clean — a human MR note that doesn't create an unresolved discussion thread would be missed. Unless I'm missing something.

bot/tests/test_preflight.py
```diff
@@ -0,0 +1,298 @@
+"""Tests for preflight runner (bot/preflight.py)."""
```
florkbr, line 1: common.py (build_repo_lookup, upstream_repo, match_repo_labels, fmt_comments), gh_pr_status.py (classify_gh, has_new_feedback), gl_mr_status.py (classify_gl), and 04-find-work.py (match_repo_labels, get_candidates) could benefit from test coverage I think.

These are pure functions with real classification logic — a bug in classify_gh could cause the bot to miss failing CI or skip feedback. Non-blocking, but worth a follow-up ticket for unit tests on the shared modules.

florkbr:
FWIW overall I think this is well structured. So if we want to follow up on these comments - feel free to ship it

presets/shared/preflight/common.py
Hyperkid123: Added a `max_comments=30` param to `fmt_comments()`. When exceeded, output is truncated with a notice pointing to Jira for full history. Keeps prompt size reasonable for tickets with long comment threads.

bot/preflight.py
Hyperkid123: Agreed — changed the aggregation: errored scripts are now excluded and the remaining scripts are aggregated normally. Only when ALL scripts error does the cycle block. Error content is still included as a `[PREFLIGHT ERROR]` header in the prompt/transcript so the session knows something was degraded.

presets/shared/preflight/gl_mr_status.py
Hyperkid123: Good catch — added `has_new_feedback()` to GL with the same logic as GH (checks non-bot notes since `last_addressed`). Wired it into the bucketing as a fallback before CLEAN, matching the GH pattern.

bot/tests/test_preflight.py
Hyperkid123: Added `test_preflight_shared.py` with 33 tests covering: `upstream_repo`, `build_repo_lookup`, `is_bot_author`, `fmt_comments` (including truncation), `classify_gh`, `classify_gl`, and `has_new_feedback` for both GH and GL. All 59 tests passing.

PR #287, opened 2026-06-30
docs: Git Auth Reverse Proxy design

Hyperkid123:
## RHCLOUD-48838 — Spike: Git Auth Reverse Proxy design

Design doc for replacing the credential-helper git auth flow with an HTTP reverse proxy in the executor server.

### What this does

Adds `docs/git-auth-proxy.md` covering:
- Problem: PATs transit through workload container memory via gRPC credential helper
- Solution: HTTP reverse proxy (`gitauth.go`) in executor server injects auth on upstream leg
- Architecture: new listener on `:8447`, reuses `devbot-proxy` ClusterIP Service
- URL scheme: `http://devbot-proxy:8447/<host>/<path>` with `insteadOf` rewrites in `.gitconfig`
- Host registry: allowlist of git providers with per-host auth config
- 3-phase migration plan (add proxy → switch workload → remove credential helper)
- Security considerations and testing plan

### What this does NOT do

No code changes. Design doc only. Once approved, this spike becomes an epic with implementation tickets.

Jira: https://redhat.atlassian.net/browse/RHCLOUD-48838

### DIFF
--- a/docs/git-auth-proxy.md
+++ b/docs/git-auth-proxy.md
@@ -313,6 +313,8 @@ This is a separate follow-up — both paths can coexist during migration.
 | `TestGitAuthProxy_Healthz` | `/healthz` → 200 |
 | `TestGitAuthProxy_LargeBody` | Simulated pack push (large POST body) → forwarded correctly |
 | `TestGitAuthProxy_MissingToken` | Host configured but env var empty → 503 |
+| `TestGitAuthProxy_EncodedHost` | `/%67ithub.com/org/repo` or `/github.com%2F..%2Fevil.com/...` → 403 (not in registry after decode) |
+| `TestGitAuthProxy_PathTraversal` | `/github.com/../evil.com/org/repo` → 403 (cleaned path doesn't match registry) |
 
 ### Integration Tests
 
--- a/docs/git-auth-proxy.md
+++ b/docs/git-auth-proxy.md
@@ -1,0 +1,338 @@
+# Git Auth Reverse Proxy Design
+
+HTTP reverse proxy that injects Git credentials into push/pull requests, eliminating PAT exposure in workload pods.
+
+---
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


docs/git-auth-proxy.md
```diff
@@ -0,0 +1,338 @@
...
+
+| Test | What |
+|---|---|
+| `TestGitAuthProxy_GitHub` | GitHub URL → Bearer token injected, HTTPS upstream |
+| `TestGitAuthProxy_GitLab` | GitLab URL → Basic auth injected, HTTPS upstream |
+| `TestGitAuthProxy_UnknownHost` | Unknown host → 403 |
+| `TestGitAuthProxy_NoHost` | Bare path (no host segment) → 400 |
+| `TestGitAuthProxy_Healthz` | `/healthz` → 200 |
+| `TestGitAuthProxy_LargeBody` | Simulated pack push (large POST body) → forwarded correctly |
+| `TestGitAuthProxy_MissingToken` | Host configured but env var empty → 503 |
```
florkbr, line 315: We may want to add a test for double-encoded or path-traversal-style URLs. The security section (line 268) correctly states the upstream URL is constructed from the registry entry, but a test would lock that in.

Something like:
| Test | What |
| --- | --- |
| TestGitAuthProxy_EncodedHost | /%67ithub.com/org/repo or /github.com%2F..%2Fevil.com/... → 403 (not in registry after decode) |

florkbr (approved):
One small request, otherwise :shipit:

docs/git-auth-proxy.md
Hyperkid123: Good call — added `TestGitAuthProxy_EncodedHost` and `TestGitAuthProxy_PathTraversal` to the test plan. Both should 403 after URL decoding/cleaning since the resulting host won't match the registry.

PR #312, opened 2026-07-15
fix(slack): support Incoming Webhook payload format

Jakub007d:
## Summary
- Auto-detect webhook type from URL: `{"text": ...}` for Incoming Webhooks (`/services/`), `{"msg": ...}` for Workflow Builder (`/triggers/`)
- Backward compatible — existing Workflow Builder webhooks continue to work unchanged
- Docs updated in `envs.md`, `workflows.md`, `onboarding-new-instance.md` to document both webhook types and recommend Incoming Webhooks

## Why
Slack Workflow Builder's "Send a message" step wraps variable content in `rich_text` blocks, which means mrkdwn features like `<url|label>` hyperlinks and `<!subteam^ID>` mentions don't render. Incoming Webhooks (`hooks.slack.com/services/`) support full mrkdwn, making notifications richer and more useful.

## Changes
- `memory-server/bot_memory_server/tools/slack.py` — detect `/services/` in the webhook URL and use `{"text": message}` payload; fall back to `{"msg": message}` for Workflow Builder URLs
- `docs/presets/envs.md` — documented both webhook types with formatting capabilities
- `docs/presets/workflows.md` — updated `SLACK_WEBHOOK_URL` description
- `docs/onboarding-new-instance.md` — updated env var table

## Test plan
- [ ] Test with Incoming Webhook URL (`hooks.slack.com/services/...`) — message delivered with full mrkdwn
- [ ] Test with Workflow Builder URL (`hooks.slack.com/triggers/...`) — existing behavior unchanged

### DIFF
--- a/memory-server/bot_memory_server/tools/slack.py
+++ b/memory-server/bot_memory_server/tools/slack.py
@@ -54,8 +54,12 @@ def register_slack_tools(mcp: FastMCP):
             }
 
         try:
+            if "/services/" in webhook_url:
+                payload = {"text": message}
+            else:
+                payload = {"msg": message}
             async with httpx.AsyncClient(timeout=10) as client:
-                resp = await client.post(webhook_url, json={"msg": message})
+                resp = await client.post(webhook_url, json=payload)
                 resp.raise_for_status()
         except Exception as e:
             logger.error("Slack webhook failed: %s", e)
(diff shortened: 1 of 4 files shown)


memory-server/bot_memory_server/tools/slack.py
```diff
@@ -54,8 +54,12 @@ async def slack_notify(
             }
 
         try:
+            if "/services/" in webhook_url:
```
Hyperkid123, line 57: I don't think we want any hardcoded values here. Can the payload be the entire JSON, and that would be decided by the skill/agent?

We will also require a faillback to ensure some migration window is active for both variants.

Hyperkid123 (dismissed):

Hyperkid123 (approved):

PR #331, opened 2026-07-21
feat(tasks): add pause/unpause task endpoints, dashboard controls, and operations guide

petrsimon:
### Description

Add pause and unpause functionality for tasks, including API endpoints, dashboard UI controls, and documentation.

[RHCLOUD-49238](https://issues.redhat.com/browse/RHCLOUD-49238)

---

### Blast radius

Affects task management API and dashboard UI. Consumers using the tasks API gain two new endpoints (`/pause`, `/unpause`). No existing endpoints changed.

---

### Rollback plan

Git revert is sufficient — new endpoints are additive and do not modify existing behavior.

---

### Checklist

### AI disclosure
Assisted by: Claude Code

### DIFF
--- a/dashboard/src/components/ConfirmDialog.tsx
+++ b/dashboard/src/components/ConfirmDialog.tsx
@@ -1,0 +1,84 @@
+import { useRef, useEffect, useState } from 'react';
+
+interface Props {
+  open: boolean;
+  title: string;
+  message: string;
+  confirmLabel?: string;
+  cancelLabel?: string;
+  variant?: 'default' | 'danger';
+  inputLabel?: string;
+  inputPlaceholder?: string;
+  onConfirm: (inputValue?: string) => void;
+  onCancel: () => void;
+}
+
+export default function ConfirmDialog({
+  open,
+  title,
+  message,
+  confirmLabel = 'Confirm',
+  cancelLabel = 'Cancel',
+  variant = 'default',
+  inputLabel,
+  inputPlaceholder,
+  onConfirm,
+  onCancel,
+}: Props) {
+  const ref = useRef<HTMLDialogElement>(null);
+  const [inputValue, setInputValue] = useState('');
+
+  useEffect(() => {
+    const dialog = ref.current;
+    if (!dialog) return;
+    if (open && !dialog.open) {
+      setInputValue('');
+      dialog.showModal();
+    } else if (!open && dialog.open) {
+      dialog.close();
+    }
+  }, [open]);
+
+  const handleConfirm = () => {
+    onConfirm(inputLabel ? inputValue : undefined);
+  };
+
+  return (
+    <dialog
+      ref={ref}
+      className="confirm-dialog"
+      onClose={onCancel}
+      onClick={(e) => {
+        if (e.target === ref.current) onCancel();
+      }}
+    >
+      <div className="confirm-dialog-content">
+        <h3>{title}</h3>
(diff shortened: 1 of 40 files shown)


dashboard/src/components/ConfirmDialog.tsx
Hyperkid123, line 1: Just FYI, there is a task in progress to migrate the dashboard to PF.

memory-server/bot_memory_server/api.py
```diff
@@ -13,12 +13,51 @@
...
+    artifacts = _parse_json_field(row.get("artifacts")) or []
+    for artifact in artifacts:
+        if artifact.get("type") in ("pull_request", "merge_request"):
+            return "pr_open"
+    if meta.get("prs"):
+        return "pr_open"
+    return "in_progress"
+
+
+async def _active_count(pool, instance_id: str | None) -> int:
```
Hyperkid123: I wonder if we want to ignore the active limit when resuming tasks?

memory-server/bot_memory_server/api.py
```diff
@@ -125,27 +164,139 @@ async def api_task_delete(request: Request) -> JSONResponse:
...
+        " UPDATE tasks SET status = $2::task_status, paused_reason = NULL,"
+        " metadata = metadata - 'status_before_pause'"
+        " FROM capacity"
+        " WHERE tasks.external_key = $1 AND tasks.status = 'archived'::task_status"
+        " AND capacity.cnt < $4"
+        " RETURNING tasks.*",
         key,
+        new_status,
+        list(ACTIVE_STATUSES),
+        MAX_ACTIVE,
```
Hyperkid123: Yeah, wondering if we want to ignore the limit. When restoring status, it usually means the priority should be up there. The task was worked on but stopped for some reason. I don't think we want to punt it all the way down in the priority list. If anything, it should be max priority IMO.

dashboard/src/components/ConfirmDialog.tsx
petrsimon: Hope it's ok I've added the light theme :) I'm one of those never went to the dark side of the Force.

petrsimon: Noted, thanks for the heads up! This is intentionally minimal since it'll be replaced during the PF migration.

memory-server/bot_memory_server/api.py
petrsimon: Good point — removed the capacity check from both unpause and unarchive. Resumed tasks were already active before, so the limit shouldn't block them.

memory-server/bot_memory_server/api.py
petrsimon: Agreed — removed the CTE capacity gate here. The queries are now simple UPDATEs. The `_active_count` helper is also gone since nothing calls it anymore.

petrsimon:
> /ok-to-test

gateway timeout :'( https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/hcc-platex-services-tenant/pipelinerun/platform-frontend-ai-dev-memory-server-on-pull-request-vdgpr/logs/run-migration-tests

I've been told to be patient and keep trying later 🤞 

will try to reproduce locally in the mean time...

passes locally, waiting for clues from Konflux, when someone replaces batteries there 😁

Hyperkid123 (approved):
LGTM

PR #340, opened 2026-07-23
feat: add Go 1.26.5 to container GOVERSIONS

lpichler:
## Summary
- Add Go 1.26.5 to the `GOVERSIONS` build arg in the Dockerfile
- Required by `entitlements-api-go` (its `go.mod` specifies `go 1.26.5`)
- Bot switches at runtime with `eval "$(use-go 1.26.5)"`

RHCLOUD-49557

## Test plan
- [ ] Verify container builds with Go 1.26.5 installed
- [ ] Verify `use-go 1.26.5` works at runtime

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,7 @@ RUN ln -sf /usr/bin/python3.12 /usr/bin/python3 \
 
 # Go — multiple versions via GOVERSIONS build arg
 # Default Go is the first version listed. Bot switches with: eval "$(use-go 1.25.7)"
-ARG GOVERSIONS="1.24.2 1.25.7"
+ARG GOVERSIONS="1.24.2 1.25.7 1.26.5"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -62,7 +62,7 @@ RUN ln -sf /usr/bin/python3.12 /usr/bin/python3 \
 
 # Go — multiple versions via GOVERSIONS build arg
 # Default Go is the first version listed. Bot switches with: eval "$(use-go 1.25.7)"
-ARG GOVERSIONS="1.24.2 1.25.7"
+ARG GOVERSIONS="1.24.2 1.25.7 1.26.5"
```
Hyperkid123, line 65: This is not for the runners. We have dockerfile.runner. Plus you should be able to install these directly in your setup.sh. We do not want to govern versions. We only added the goverion cli so you can do this.

PR #343, opened 2026-07-24
Add queue deduplication and digest hour timing with deterministic coo…

radekkaluzik:
…ldown key

### Description

Follow-up to rehor#316 addressing review feedback:

* **Queue deduplication** — `slack_notify` in digest mode checks for existing unsent `(jira_key, event_type)` before inserting, preventing duplicate entries when the bot processes the same ticket across multiple cycles
* **Runner-triggered digest (zero tokens)** — `bot/run.py` calls `bot/slack_digest.py` directly after each cycle (not as subprocess — direct Python import). Timing (hour check, weekend skip) is handled in the script, not in the memory server tool — each instance controls its own schedule via `SLACK_DIGEST_HOUR` env var (default: 9). No LLM involvement, no CronCreate needed.
* **Reliable hour matching** — uses `now.hour >= digest_hour` instead of exact match, with sent-today deduplication via deterministic `digest_key` in `slack_notifications` table. Digest fires even if a cycle misses the exact hour.
* **Simplified `slack_send_digest` tool** — removed `digest_hour` parameter, hour/weekend checks. Tool handles queue processing and sent-today deduplication via `digest_key`. All timing logic lives on the instance side.
* **Removed `slack-digest` skill** — digest is not a skill anymore, it's a bot runner module (`bot/slack_digest.py`). Agent doesn't need to know about digest mode. Removed SKILL.md, skill directories, manifest registration, and CLAUDE.md references.
* **Moved `memory_mcp` to `bot/`** — proper module import instead of fragile `sys.path.insert` reaching into `.claude/skills/`
* **Testable without SDK** — `try_slack_digest()` lives in `bot/slack_digest.py`, tests import it directly instead of copying the function

[RHCLOUD-48014](https://redhat.atlassian.net/browse/RHCLOUD-48014)

---

### Blast radius

* **Bot runner** (`bot/run.py`) — new `_try_slack_digest()` called after each cycle, imports `try_slack_digest` from `bot.slack_digest`
* **Memory server** — `slack_send_digest` accepts optional `digest_key` for sent-today deduplication. Backward compatible.
* **New module** — `bot/memory_mcp.py` (copy of `.claude/skills/memory_mcp.py`, used by `slack_digest.py`)
* **Removed** — `.claude/skills/slack-digest/` and `presets/envs/slack/skills/slack-digest/` directories, `slack-digest` from slack env manifest
* No changes to immediate mode behavior

---

### Rollback plan

Git revert is sufficient. Digest queue and notifications table are unchanged structurally.

---

### Checklist

### AI disclosure


Hyperkid123:
@radekkaluzik 

>Memory server — slack_send_digest has a new optional digest_hour parameter (default 9, backward compatible)

This can't be part of the memory server but of the agent runners. Each instance sets this on its own. I also think the digest does not have to be triggered by the agent at all. We can trigger it, or try to trigger it before/after each cycle, provided that the conditions are met. In this case, the skill/script exists, the Slack webhook exists, and the digest hour was set. I think we can get this down to 0 tokens spent.

radekkaluzik:
> @radekkaluzik
> 
> > Memory server — slack_send_digest has a new optional digest_hour parameter (default 9, backward compatible)
> 
> This can't be part of the memory server but of the agent runners. Each instance sets this on its own. I also think the digest does not have to be triggered by the agent at all. We can trigger it, or try to trigger it before/after each cycle, provided that the conditions are met. In this case, the skill/script exists, the Slack webhook exists, and the digest hour was set. I think we can get this down to 0 tokens spent.

ok, makes sense. I'll implement that.

presets/core/CLAUDE.md
```diff
@@ -161,12 +161,7 @@ python3 .claude/skills/slack-notify/slack_notify.py <JIRA_KEY> <EVENT_TYPE> "<ME
 
 Script reads `$SLACK_WEBHOOK_URL` from env. No webhook → silent no-op. 48h cooldown per external_key (automatic).
 
-**Daily digest mode** (`SLACK_NOTIFY_MODE=daily_digest`): Events queue instead of sending immediately. Trigger digest with `/slack-digest digest` or schedule via CronCreate:
-
-```bash
-python3 .claude/skills/slack-digest/slack_cmd.py digest 2>&1
-python3 .claude/skills/slack-digest/slack_cmd.py status <JIRA_KEY> 2>&1
-```
+**Daily digest mode** (`SLACK_NOTIFY_MODE=daily_digest`): Events queue instead of sending immediately. The bot runner automatically triggers the digest at the configured hour — no agent action needed.
```
Hyperkid123: I don't think that Claude needs to know anything about Daily Digest now. It does not trigger it anymore.

bot/run.py
```diff
@@ -212,6 +208,36 @@ def _check_wake_signal(instance_id: str) -> bool:
...
+    logger = logging.getLogger(__name__)
+    digest_script = Path(__file__).resolve().parent / "slack_digest.py"
+    if not digest_script.exists():
+        return
+    if not os.environ.get("SLACK_WEBHOOK_URL"):
+        return
+
+    try:
+        result = subprocess.run(
+            [sys.executable, str(digest_script), "digest"],
```
Hyperkid123: Do we need to run a subprocess? The digest is jsut a pyton. can't we just import it? And run as regular python code?

radekkaluzik: Fixed

presets/core/CLAUDE.md
radekkaluzik: Fixed

bot/slack_digest.py
```diff
@@ -1,18 +1,21 @@
 #!/usr/bin/env python3
-"""Slack notification management -- digest and status updates.
+"""Slack daily digest — runner-triggered, zero LLM tokens.
+
+Called by bot/run.py after each cycle. Checks conditions locally
+(hour, weekend, webhook) before calling the memory server MCP tool.
 
 Usage:
```
Hyperkid123, line 7: `sys.path.insert(0, ... / ".claude" / "skills")` — this file moved to `bot/` but still reaches back into `.claude/skills/` for imports. Fragile path that breaks if skill directory changes or isn't installed yet. Move `memory_mcp` into `bot/` as a proper module or add it as a package dependency.

bot/tests/test_slack_digest_runner.py
```diff
@@ -0,0 +1,46 @@
...
+replicate the function logic and test it independently.
+"""
+
+import os
+from unittest.mock import patch
+
+
+def _try_slack_digest() -> None:
+    """Copy of bot.run._try_slack_digest for testing without claude_agent_sdk."""
+    if not os.environ.get("SLACK_WEBHOOK_URL"):
```
Hyperkid123: This test copies `_try_slack_digest()` instead of importing the real function. If the real implementation in `bot/run.py` changes, this test still passes — it tests a stale copy, not actual code. Either mock `claude_agent_sdk` so you can import `bot.run`, or restructure so the testable logic lives in a module that doesn't pull in the SDK.

bot/slack_digest.py
```diff
@@ -1,18 +1,21 @@
...
 """
 
 import json
 import os
 import sys
+from datetime import datetime, timezone
 from pathlib import Path
 
-sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / ".claude" / "skills"))
```
Hyperkid123: `now.hour != digest_hour` — exact hour match is unreliable. If a cycle runs at 8:59 and the next at 10:01, digest never fires that day. Consider tracking "already sent today" (e.g. a flag file or DB check with date) and sending if `now.hour >= digest_hour` and not yet sent today.

memory-server/tests/test_slack_digest.py
```diff
@@ -286,39 +302,20 @@ async def test_send_digest_success(self, slack_tools):
         assert "RHCLOUD-100" in payload["msg"]
         assert "RHCLOUD-101" in payload["msg"]
 
-        update_call = pool.execute.call_args
-        assert "UPDATE" in update_call[0][0]
-        assert update_call[0][1] == [1, 2]
-
     @pytest.mark.asyncio
```
Hyperkid123, line 305: `test_send_digest_idempotent` was removed but it tested valid behavior — second call after all items sent returns empty. This is still relevant since `slack_send_digest` is still the MCP tool that processes the queue. Consider keeping it.

bot/slack_digest.py
radekkaluzik: Fixed

bot/tests/test_slack_digest_runner.py
radekkaluzik: Fixed

bot/slack_digest.py
radekkaluzik: Fixed

memory-server/tests/test_slack_digest.py
radekkaluzik: Fixed

Hyperkid123 (approved):

PR #381, opened 2026-07-30
fix(onboarding): fix preflight not responding to JIRA comments

florkbr:
PR 1 — Onboarding preflight fix:

  ### Description

  Fix onboarding preflight comment detection — the MCP `jira_get_issue` call in `_jira_issue()` omitted `comment` from the `fields` parameter, causing the mcp-atlassian tool to never return comments. This meant
  `_has_new_jira_feedback()` always received an empty list and returned `False`, so the preflight never detected user replies on onboarding tickets. The bot silently skipped every human response.

  **How we verified this is a real bug:**

  1. Observed the onboarding bot ignoring a human comment on REHOR-63 — preflight consistently returned `skip` even after a direct reply to the bot's intake questions.
  2. Called `jira_get_issue` with the same params the preflight used (`fields="summary,status,assignee,labels,issuelinks"`, `comment_limit=10`) — response contained **no** `comments` key. Added `comment` to `fields` — comments
  appeared at the top-level `comments` key. The `comment_limit` parameter alone is insufficient; `comment` must be in `fields`.
  3. Traced the code path: `_jira_issue()` → empty `comments` → `_has_new_jira_feedback()` returns `False` → preflight emits `skip` → bot never wakes up for user feedback.
  4. Confirmed this is a bug from inception (commit `28819a6`, July 2) — `comment` was never included in any preflight's `_jira_issue` fields parameter.

  Adds 21 unit tests for the onboarding preflight covering feedback detection, blocked tasks, onboarding label parsing, candidate discovery, auto-advance, error handling, and Jira unavailability.

  ---

  ### Blast radius
  
  - **Onboarding workflow only** — only `onboarding_preflight.py` is modified.
  - No changes to sprint/kanban/triage preflights (those are a separate PR).
  - The rehor-onboarding-agent instance is the only consumer of this preflight.

  Tested by:
  - Empirically verifying MCP returns comments with the fix applied
  - 21 new unit tests + full bot test suite (267 tests pass)
  - ruff 0.16.0 format, lint, and check clean
  - This changes skip/start behavior: tickets with new human comments that previously produced `skip` will now produce `start`, triggering an LLM cycle. This is the intended behavior but will increase token usage for instances
  that receive frequent Jira comments.

  ---

  ### Rollback plan

  Git revert is sufficient — reverting removes `comment` from fields, returning to the (broken) previous behavior. No data migrations or state changes.

  ---

  ### Checklist

  ### AI disclosure

### DIFF
--- a/bot/tests/test_onboarding_preflight.py
+++ b/bot/tests/test_onboarding_preflight.py
@@ -1,0 +1,333 @@
+"""Tests for onboarding_preflight module."""
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+SHARED_DIR = Path(__file__).resolve().parent.parent.parent / "presets" / "shared" / "preflight"
+SKILLS_DIR = Path(__file__).resolve().parent.parent.parent / ".claude" / "skills"
+sys.path.insert(0, str(SHARED_DIR))
+sys.path.insert(0, str(SKILLS_DIR))
+
+from onboarding_preflight import (  # noqa: E402
+    _get_onboarding_label,
+    _has_new_jira_feedback,
+    _is_blocked,
+    main,
+)
+
+# --- _has_new_jira_feedback ---
+
+
+def test_feedback_detected_with_comments_field():
+    """The fix: when MCP returns comments at top-level, feedback is detected."""
+    issue = {
+        "comments": [
+            {
+                "created": "2026-07-01T10:00:00",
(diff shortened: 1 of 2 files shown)


florkbr:
Example of bot failing to acknowledge a JIRA comment: https://redhat.atlassian.net/browse/REHOR-63?focusedCommentId=17803362

PR #415, opened 2026-08-06
REHOR-107: container build verification workflow and smoke checks

vishsanghishetty:
Jira: https://redhat.atlassian.net/browse/REHOR-107

## Summary

Konflux already builds the bot and memory-server images on every PR (and proxy when `proxy/**` changes), but nothing verifies a built image actually starts. Adds `.github/workflows/container-verify.yml` with one job per image so a failure is unambiguous in the PR checks list:

- `verify-bot`: builds the image and checks required tooling is present (`python3`, `uv`, `git`, `tini`, `bwrap`, `buildah`, `node`, `go`, `gh`, `glab`, `gpg`, etc). `entrypoint.sh` hard-waits up to 30s on a Unix socket produced by the proxy sidecar's executor-server and exits if it's not there, so a standalone run of the real entrypoint always dead-ends. This bypasses the entrypoint and verifies the build itself instead — full entrypoint-driven verification is REHOR-62's scope (documented below).
- `verify-proxy`: builds the image, starts it with a dummy `GH_TOKEN`, and checks Squid answers `mgr:info` and the executor socket exists. `start.sh` only requires GitLab/GPG/Jira secrets conditionally, so this is a real functional smoke check, not just a build check.
- `verify-memory-server`: builds the image, starts it against a real `pgvector/pgvector:pg17` service, and polls `/health`.

Also:
- Adds `docs/operations/rehor-107-container-verification.md` documenting what each job verifies and the REHOR-62 hand-off for full multi-container runtime verification
- Adds informational (non-required, path-filtered) entries for the three new jobs to `.github/branch-protection/required-checks.json` — no branch protection or admin changes, REHOR-77 already closed that loop
- Updates `docs/roadmap.md` CI Foundation status
- Adds a `make container-verify` local parity target
- Fixes a pre-existing `CONTAINER_RT` bug in the `Makefile` (`command -v docker` was printing its path to stdout in addition to the intended `echo`, so the variable resolved to `"docker docker"`), which silently broke `make memory-server`, `make docker-up`, and `make docker-down` — unrelated to REHOR-107 but directly blocked validating the new target locally

## Testing done

- `make verify` (full existing suite): passed — Python 818 passed / 2 xfailed / 1 xpassed, Go vet clean + tests `ok`, Dashboard typecheck/build/61 tests passed. No regression to existing checks.
- `make container-verify` (new target), run multiple times end-to-end through the real Makefile target:
  - `verify-bot`: image builds; all 13 required tools present and runnable — confirmed 3x
  - `verify-proxy`: image builds; Squid answers `mgr:info`; executor socket created — confirmed 3x
  - `verify-memory-server`: image builds; connects to real pgvector Postgres; reaches "Memory server ready" and serves on 8080 — confirmed 4x. (Running all three sequentially in one invocation intermittently hits a local Docker Desktop VM disk ceiling on my machine after the first two heavy builds — isolated and confirmed this is a local-environment limitation, not a logic defect; each job passes cleanly on its own with a clean disk slate, and GitHub Actions runners get a fresh disk per job.)
- Verified the workflow only triggers on relevant path changes (bot/proxy/memory-server image inputs), not on unrelated file changes
- Workflow YAML and `required-checks.json` syntax validated (`yaml.safe_load` / `json.load`)
- Pre-commit and pre-push hooks passed on commit/push

## Validation

- [x] `make verify` passes
- [x] `make container-verify` — each job individually confirmed passing (see testing notes above on the local disk-ceiling caveat)
- [x] Workflow path-filter behavior confirmed
- [ ] Live PR check run on GitHub Actions (will confirm once this PR is open)

### DIFF
--- a/.github/workflows/container-verify.yml
+++ b/.github/workflows/container-verify.yml
@@ -1,0 +1,204 @@
+name: Container Verify
+
+# Builds the bot, proxy, and memory-server images and runs a minimal smoke
+# check against each. This is intentionally path-filtered like every other
+# GitHub Actions workflow in this repo: GitHub does not render a "skipped"
+# badge for a non-triggered path-filtered workflow, it simply doesn't appear
+# on the PR. "Non-blocking for unrelated PRs" means exactly that: unrelated
+# file changes never surface this check at all.
+#
+# Full multi-container runtime verification (executor sidecar + proxy +
+# memory-server + real entrypoint stages) is out of scope here — see
+# docs/operations/rehor-107-container-verification.md for the REHOR-62
+# hand-off notes.
+
+on:
+  pull_request:
+    paths:
+      # bot image inputs
+      - "Dockerfile"
+      - "entrypoint.sh"
+      - "config.json"
+      - ".mcp.json"
+      - "CLAUDE.md"
+      - ".claude/**"
+      - "presets/**"
+      - "dev-proxy/**"
+      - "bot/**"
+      - "pyproject.toml"
+      - "uv.lock"
+      # proxy image inputs
+      - "proxy/**"
+      # memory-server image inputs
+      - "memory-server/**"
+      # shared
+      - "docker-compose.yml"
+      - ".github/workflows/container-verify.yml"
+  push:
+    branches:
+      - master
+    paths:
+      # bot image inputs
+      - "Dockerfile"
+      - "entrypoint.sh"
+      - "config.json"
+      - ".mcp.json"
+      - "CLAUDE.md"
+      - ".claude/**"
+      - "presets/**"
+      - "dev-proxy/**"
+      - "bot/**"
+      - "pyproject.toml"
+      - "uv.lock"
+      # proxy image inputs
+      - "proxy/**"
+      # memory-server image inputs
+      - "memory-server/**"
+      # shared
+      - "docker-compose.yml"
+      - ".github/workflows/container-verify.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  verify-bot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+
+      - name: Build bot image
+        run: |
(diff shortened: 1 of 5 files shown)


vishsanghishetty:
@Hyperkid123 ready for review

.github/workflows/container-verify.yml
```diff
@@ -0,0 +1,211 @@
...
+  verify-bot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+
+      - name: Build bot image
+        run: |
+          docker build --build-arg GOVERSIONS="1.24.2 1.25.7" -t bot:verify -f Dockerfile .
```
Hyperkid123: The agents builds are using the `Dockerfile.runner`file, we should be using the same

.github/workflows/container-verify.yml
```diff
@@ -0,0 +1,211 @@
...
+              break
+            fi
+            sleep 2
+          done
+          if [ "$squid_ok" -ne 1 ]; then
+            echo "::error::proxy image built but squid did not become ready within 30s"
+            docker logs proxy-verify
+            exit 1
+          fi
+          echo "squid is ready"
```
Hyperkid123, line 162: We probably want to check if our network filter is working. Try to make some requests that should pass/fail. I know it's trivial but just in case squid falls apart.

.github/workflows/container-verify.yml
```diff
@@ -0,0 +1,211 @@
...
+          if ! docker exec proxy-verify test -S /var/run/devbot/executor.sock; then
+            echo "::error::proxy image built but executor-server did not create its listening socket"
+            docker logs proxy-verify
+            exit 1
+          fi
+          echo "executor socket is present"
+
+      - name: Cleanup
+        if: always()
+        run: docker rm -f proxy-verify
```
Hyperkid123, line 175: Same here; probably should be a follow-up. Try to fire some GRCP requests and test that the blocked cli commands on GH are properly handled

Hyperkid123 (commented):
Just on real thing to adress, rest should be in the 62 followup

Hyperkid123 (approved):

PR #420, opened 2026-08-07
feat: add Prometheus operational metrics (REHOR-46)

petrsimon:
## Summary

- Add `prometheus-client` to bot with 11 operational metrics served on `:9091`:
  - Counters: preflight outcomes, config sync, cycle timeouts, work types, turn budget events, transcript uploads, MCP server status, wake signals
  - Histogram: cycle duration (wall-clock per cycle with latency distribution buckets)
  - Gauges: preflight consecutive errors, disk free MB
- Wire two new proxy `CounterVec`s (`executor_policy_deny_total`, `vertex_model_requests_total`) into existing `:9091` metrics server
- Vertex proxy metric uses actual HTTP status code from upstream (not hardcoded "ok")
- Add bot Service, container port declaration, and monitoring NetworkPolicy to the instance deploy template
- Bot uses same metrics port (`:9091`) as proxy — separate pods, no conflict — so existing ServiceMonitors pick up bot metrics automatically
- Cost/token metrics are handled separately in REHOR-45 (memory-server side)

[REHOR-46](https://issues.redhat.com/browse/REHOR-46)

---

### Blast radius

- **Bot pods**: new `:9091` metrics HTTP server starts at boot — no impact on existing bot behavior, metrics are additive
- **Proxy**: two new counters registered in existing Prometheus registry — no behavioral change
- **Deploy template** (`deploy-template.yaml.j2`): adds a new Service and NetworkPolicy per bot instance — existing resources untouched

---

### Rollback plan

`git revert` is sufficient — metrics are purely additive with no schema migrations or external state.

---

### Checklist

### AI disclosure
Assisted by: Claude Code


bot/metrics.py
Hyperkid123, line 1: I'd like to see a metric that also outputs token/cycle. We have this information. The thing I am most interested in is if there are any cycles marked as idle with consumed tokens. If there are, it means a bug in a preflight check. We should also output what workflow was used. We should have all this information in the agent pods.

bot/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+    "devbot_transcript_upload_total",
+    "Transcript upload outcomes.",
+    ["label", "outcome"],
+)
+MCP_SERVER_STATUS_TOTAL = Counter(
+    "devbot_mcp_server_status_total",
+    "MCP server connection status observed at cycle start.",
+    ["server", "status"],
+)
+WAKE_SIGNAL_TOTAL = Counter(
```
Hyperkid123: I don't think this works. we should not log it and remove the task at some point. Now that long sleep is not a thing we should no longer need it.

bot/transcripts.py
```diff
@@ -78,6 +79,7 @@ def record_transcript(
     session_id = getattr(result, "session_id", "")
     if not session_id:
         logger.debug("No session_id in result — skipping transcript capture")
+        TRANSCRIPT_UPLOAD_TOTAL.labels(label, "no_session").inc()
```
Hyperkid123: I think we also want to if no transcript was uploaded for a cycle. The only acceptable occurrence of this is when a preflight does not trigger a session. If a cycle ended with no transcript, we have an issue, and we should send an alert.

bot/metrics.py
petrsimon: Done on the agent side. We now emit real-time `devbot_cycle_*` cost/token counters with a `workflow` label, plus `devbot_idle_with_tokens_total` for idle cycles that still spent tokens (preflight bug signal). Wired from `record_cost()` where usage and `no_work` already live. Memory-server keeps only DB gauges (#421).

bot/metrics.py
petrsimon: Agreed. Dropped `WAKE_SIGNAL_TOTAL` and removed bot wake polling — sleep is a plain blocking `time.sleep` again. Filed [REHOR-118](https://issues.redhat.com/browse/REHOR-118) to clean up the leftover memory-server `/wake` API and dashboard buttons (now no-ops).

bot/transcripts.py
petrsimon: Done. Session cycles without a transcript now increment `devbot_transcript_upload_total{outcome="missing"}` (alertable). `ok` only when a transcript was actually attached. Preflight skip/error still go through `post_orphan_cycle` and don’t hit this path.

Hyperkid123 (approved):

PR #421, opened 2026-08-07
feat(memory-server): add custom Prometheus cost/token metrics

petrsimon:
## Summary

- Add real-time Prometheus counters (`devbot_cycle_*`) incremented on every `POST /api/costs` — tracks cost USD, input/output/cache tokens, cycle count by status
- Add DB-backed gauges (`devbot_db_*`) refreshed from Postgres every 60s for reconciliation
- Add test coverage for `record_cycle` (None coercion, accumulation, defaults) and `refresh_db_gauges`

[REHOR-45](https://issues.redhat.com/browse/REHOR-45)

---

### Description

Two layers of metrics for cycle cost tracking in the memory server:

1. **Real-time counters** — incremented synchronously in the `/api/costs` POST handler via `record_cycle()`. Labels: `model`, `label`, `status` (ok/error/idle).
2. **DB-backed gauges** — a background asyncio task queries `cycles` aggregate totals every 60s and sets gauge values. Useful for dashboard reconciliation against counters that reset on restart.

Duration is intentionally omitted — [REHOR-46](https://issues.redhat.com/browse/REHOR-46) tracks `devbot_cycle_duration_seconds` as a histogram in the bot process (percentiles > averages).

Also renames `api_costs_add` → `_api_costs_add` (private helper, not directly routed).

---

### Blast radius

- `memory-server` only — metrics exposed on existing `:9091/metrics` endpoint
- No changes to existing metrics or API contracts
- New metrics are additive; Grafana dashboards can be wired up separately

---

### Rollback plan

- Git revert is sufficient — no schema changes, no external dependencies

---

### Checklist

### AI disclosure
Assisted by: Claude Code


memory-server/bot_memory_server/metrics.py
```diff
@@ -17,6 +23,137 @@
...
+    input_tokens: int,
+    output_tokens: int,
+    cache_read_tokens: int,
+    cache_write_tokens: int,
+) -> None:
+    """Increment real-time cycle counters. Call on every POST /api/costs."""
+    model = model or "unknown"
+    label = label or "unknown"
+
+    CYCLE_COST_USD_TOTAL.labels(model=model, label=label).inc(cost_usd or 0)
```
Hyperkid123: This may be overlapping with the agent metrics, but we want to know when a cycle was reported as idle, yet tokens were spent. All idle cycles should be caught by preflight. Maybe we will have to start sending the preflight result as well to this to get more accurate metrics, or we can send it fomr the agent pod directly sicne we have all the information there.

That brings me to a question. Do we need these metrics here? Can'r we send it all from the agent pods? They already have all this information, we might as well report it form the source.

petrsimon: Agreed — real-time cost/token counters don’t belong in memory-server.

Now it's like this:

- Agent (REHOR-46 / #420): real-time devbot_cycle_* counters, including idle-with-tokens + workflow, where preflight and usage already live.
- Memory-server (this PR): keep only the DB-backed devbot_db_* gauges for fleet reconciliation across restarts.

Real-time counters and record_cycle on POST /api/costs are removed here. Idle+tokens anomaly detection will come from the agent metrics, not by piping preflight into memory-server.

Hyperkid123 (approved):

PR #423, opened 2026-08-10
feat(e2e): add E2E testing infrastructure support (REHOR-110)

Hyperkid123:
## Summary

- **Playwright browser alignment**: `align-playwright-browsers` helper in `browser/install.sh` symlinks pre-installed browsers to match repo-pinned playwright version, avoiding runtime download
- **Credential mapping**: `10-chromium.sh` maps `SSO_USERNAME`/`SSO_PASSWORD` to `E2E_USER`/`E2E_PASSWORD` when not already set
- **Proxy allowlist**: `squid.conf` now permits `cdn.playwright.dev` and `playwright.download.prss.microsoft.com` for browser downloads
- **Dev-proxy hardening**: `dev-proxy/install.sh` creates `/etc/caddy` before copy, adds success/warning logging

## Test plan

- [x] 14 new tests in `bot/tests/test_e2e_infra.py` — all passing
- [x] Full bot test suite (323 tests) green
- [ ] Verify in framework instance by adding `dev-proxy` to envs list in `instance.yaml`
- [ ] Create `extra-hosts` file in framework instance config


presets/envs/browser/entrypoint.d/10-chromium.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 # Start headless Chromium for chrome-devtools MCP
 
+# Map SSO credentials to E2E vars (used by Playwright global-setup)
+if [ -n "${SSO_USERNAME:-}" ] && [ -z "${E2E_USER:-}" ]; then
```
vishsanghishetty: this `SSO_* -> E2E_*` mapping looks broken in the real startup flow. In `entrypoint.sh` lines 35-40 we write `.credentials` and then unset SSO_USERNAME/SSO_PASSWORD before entrypoint.d runs, and entrypoint.d scripts are executed as subprocesses at lines 195 and 198 (bash "$script"), so exports here don’t persist to the parent process that does exec at line 204. can you move this to a path that survives startup (for example, reading from `.credentials` where PW setup actually runs)?

bot/tests/test_e2e_infra.py
```diff
@@ -0,0 +1,330 @@
...
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+class TestAlignPlaywrightBrowsers:
+    """Test the align-playwright-browsers helper embedded in browser/install.sh."""
+
+    @pytest.fixture
+    def align_script(self, tmp_path):
+        """Extract align-playwright-browsers logic with portable grep (no -P)."""
```
vishsanghishetty: the test says it uses “portable grep (no -P)”, and lines 32-33 use sed, but the real shipped script in presets/envs/browser/install.sh lines 36-37 uses grep -oP. That means we’re testing a rewritten copy, not the exact script we ship. can we either execute the literal embedded script in tests or add a parity assertion so regressions in the real script don’t slip through?

vishsanghishetty (approved):
lgtm!

presets/envs/browser/entrypoint.d/10-chromium.sh
```diff
@@ -1,6 +1,18 @@
 #!/bin/bash
 # Start headless Chromium for chrome-devtools MCP
 
+# Map SSO credentials to E2E vars (used by Playwright global-setup).
+# SSO_USERNAME/SSO_PASSWORD are unset by entrypoint.sh before entrypoint.d runs,
+# so read from the .credentials file that entrypoint.sh writes at line 36-39.
+CRED_FILE="/home/botuser/app/.credentials"
+if [ -z "${E2E_USER:-}" ] && [ -f "$CRED_FILE" ]; then
+    E2E_USER=$(python3 -c "import json,sys; d=json.load(open('$CRED_FILE')); print(d['sso']['username'])" 2>/dev/null)
```
florkbr, line 10: These one-liners swallow errors via 2>/dev/null. If .credentials exists but has a wrong shape (e.g., missing sso.username key), E2E tests will fail later with a confusing "E2E_USER not set" error rather than pointing at the credential file. A fallback echo to stderr would help debugging.

bot/tests/test_e2e_infra.py
```diff
@@ -0,0 +1,331 @@
...
+        assert "Could not determine" in result.stderr
+
+
+class TestChromiumCredentialMapping:
+    """Test E2E credential mapping from .credentials file in 10-chromium.sh."""
+
+    @pytest.fixture
+    def credential_script(self, tmp_path):
+        """Extract the credential mapping logic that reads from .credentials."""
+        script = textwrap.dedent("""\
```
florkbr: This is building a standalone bash script in a fixture rather than extracting from `10-chromium.sh`. If someone changes the real 10-chromium.sh credential logic, these tests won't catch potential regressions. (vishali also called out here: https://github.com/OpenShift-Fleet/rehor/pull/423#discussion_r3750820488)

presets/envs/browser/install.sh
```diff
@@ -22,6 +22,45 @@ npx playwright install chromium
...
+# Usage: align-playwright-browsers <repo-dir>
+# Symlinks the installed browser dirs to the version the repo's playwright expects.
+cat > /usr/local/bin/align-playwright-browsers << 'SCRIPT'
+#!/bin/bash
+set -e
+REPO_DIR="${1:-.}"
+PW_BROWSERS="${PLAYWRIGHT_BROWSERS_PATH:-/opt/pw-browsers}"
+USER_CACHE="${HOME}/.cache/ms-playwright"
+
+WANTED=$(cd "$REPO_DIR" && npx playwright install --dry-run 2>&1 | grep -oP 'chromium-\K\d+' | head -1)
```
florkbr, line 36: @Hyperkid123 not blocking, but FYI we have a reusable NPM package for auth with Redhat SSO: https://github.com/RedHatInsights/frontend-test-utils/tree/master/packages/playwright-test-auth

Not everyone may need it obviously and we probably don't want to control the version, but may be worth exposing/documenting.

florkbr (commented):
Few comments and a bigger design Q (elephant in the room - not blocking per say but possibly worth us investigating?): is there a way to isolate the creds to just the Playwright process rather than exporting them into the entire shell session that Claude also inherits? I know these should be test users, but it would be nice to have an air-gap here as well depending on which system the bot is accessing. I've been thinking about this for other situations where a wrapper/injection is not possible.

A Kubernetes Job is the most natural fit. The bot would:
     
  1. Create a Job via the K8s API (the bot's ServiceAccount would need RBAC to create Jobs in a new namespace)
  2. The Job pod runs the Playwright tests in its own container, with SSO creds mounted as a Kubernetes Secret volume — only that pod gets the mount
  3. Bot watches the Job for completion, reads results from logs or a shared PVC/S3 artifact
  4. The main bot pod never mounts the secret at all

And by using a separate namespace:
  - The bot's ServiceAccount gets RBAC to create Jobs and read logs in the new namespace only
  - The SSO Secret lives in the new namespace — the bot's namespace never has it
  - You can set ResourceQuotas and LimitRanges on the new namespace independently (so a runaway Playwright run doesn't eat the bot's resources)
  - ResourceQuotas also give you natural E2E concurrency control — cap at the namespace level, or have each bot label its Jobs and check for an existing active run before creating another, giving you per-bot concurrency without a custom scheduler.

The tradeoff is latency — spinning up a pod takes 10-30s vs running in-process. But for E2E tests that already take minutes, that's negligible? And the job pattern may be beneficial for other use cases outside of E2E

The bot doesn't lose any debugging capability either with the Job approach, since traces/screenshots/videos are artifacts either way and interactive debugging doesn't apply in headless

Hyperkid123:
@florkbr I think what you're saying makes sense and is something I'd like to do. This is mainly for unblocking current jobs.

I agree on the high level with what you have said. Care to create a spike in https://redhat.atlassian.net/browse/REHOR-83 so we can track and implement later. Yes, these accounts should be for testing and are in an isolated env, but still we are breaking the rule a bit with these. If we solve the issue, we can prep the env for other testing scenarios.

presets/envs/browser/install.sh
Hyperkid123: @florkbr I am aware, but it is specific to your agents/env. Possibly a good candidate for a custom env in your agent(s)

presets/envs/browser/entrypoint.d/10-chromium.sh
Hyperkid123: Should be fixed now

florkbr (approved):

florkbr:
@Hyperkid123 done! https://redhat.atlassian.net/browse/REHOR-124

PR #426, opened 2026-08-11
Call _try_slack_digest in skip and error preflight paths so digest fi…

radekkaluzik:
…res even on idle cycles

### Description

`_try_slack_digest()` was only called after a completed agent cycle. When all preflight scripts return "skip" (the most common case — no work to do), the bot goes straight to sleep via `continue` and never reaches the digest call. This means digest only fires on days when the bot actively works, which defeats the purpose of a daily summary.

Fix: call `_try_slack_digest()` in all three paths — after error preflight, after skip preflight, and after agent cycle. Empty queue is a no-op (memory server returns "No items to digest"), so calling it every iteration is safe.

[RHCLOUD-48014](https://redhat.atlassian.net/browse/RHCLOUD-48014)

---

### Blast radius

* **Bot runner** (`bot/run.py`) — two additional `_try_slack_digest()` calls in existing preflight skip/error paths
* No new dependencies, no API changes
* Instances without `SLACK_DIGEST_HOUR` set are unaffected (opt-in check in `slack_digest.py`)

---

### Rollback plan

Git revert is sufficient. Digest stops firing on idle cycles, reverts to previous behavior.

---

### Checklist

### AI disclosure

### DIFF
--- a/bot/run.py
+++ b/bot/run.py
@@ -468,6 +468,8 @@ def main() -> None:
 
     try:
         while True:
+            _try_slack_digest()
+
             remote_agent_dir, shared_agent_dir = sync_config_repo()
             if shared_agent_dir:
                 apply_merged_config(SCRIPT_DIR, shared_agent_dir)
@@ -564,8 +566,6 @@ def main() -> None:
                 )
             else:
                 logger.warning("Cycle produced no result")
-
-            _try_slack_digest()
 
             _read_sleep_signal(config, instance_id)
 


petrsimon:
@Hyperkid123 I can't approve :) You can.

Hyperkid123:
@radekkaluzik do should we call this before cycle (preflight) runs? Call it in one place instead of 3?

radekkaluzik:
> @radekkaluzik do should we call this before cycle (preflight) runs? Call it in one place instead of 3?

Fixed

Hyperkid123 (approved):

PR #428, opened 2026-08-12
feat: exit after single cycle for source=run_and_exit workflows

catastrophe-brandon:
## Summary

- When `instance.yaml` sets `source: scheduled`, the runner now exits after completing one iteration instead of sleeping and looping
- This applies to all three preflight outcomes: `start` (cycle ran), `skip` (nothing to do), and `error`
- Lets KEDA be the sole source of truth for scheduling -- the pod runs once and exits, KEDA scales it back to zero, and brings it up again at the next scheduled window

Previously, scheduled workflows relied on a `cycle-sleep.json` file (24h pause) to prevent repeated runs within the same KEDA window. This was fragile across pod restarts -- when the pod came back up, the sleep state was lost and the bot would run again, wasting tokens on duplicate scans.

The `source` field was already parsed from `instance.yaml` and logged but never acted on. This change gives it operational meaning.

## Test plan

- [x] All existing 176 preflight/run tests pass
- [x] 4 new tests verify exit-after-one-iteration for all three outcomes (start, skip, error) when `source=scheduled`, and continued looping when `source=jira`
- [ ] Deploy a `source: scheduled` instance and confirm the pod exits cleanly after one cycle

### DIFF
--- a/bot/run.py
+++ b/bot/run.py
@@ -489,7 +489,7 @@ def main() -> None:
                         input_prompt=preflight_result.transcript,
                     )
                     error_sleep = min(config.interval * (2**consecutive_preflight_errors), 300)
-                    if instance_config.source == "scheduled":
+                    if instance_config.source == "keda_scheduled":
                         if consecutive_preflight_errors >= 3:
                             logger.error(
                                 "Scheduled source — giving up after %d preflight errors", consecutive_preflight_errors
@@ -525,7 +525,7 @@ def main() -> None:
                         idle_cycle_limit=instance_config.idle_cycle_limit,
                         cooldown_seconds=config.idle_reminder_cooldown_seconds,
                     )
-                    if instance_config.source == "scheduled":
+                    if instance_config.source == "keda_scheduled":
                         logger.info("Scheduled source — exiting after preflight skip")
                         cleanup_between_cycles(SCRIPT_DIR)
                         break
@@ -586,7 +586,7 @@ def main() -> None:
             else:
                 logger.warning("Cycle produced no result")
 
-            if instance_config.source == "scheduled":
+            if instance_config.source == "keda_scheduled":
                 logger.info("Scheduled source — exiting after cycle")
                 cleanup_between_cycles(SCRIPT_DIR)
                 break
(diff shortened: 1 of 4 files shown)


catastrophe-brandon:
> when the pod came back up, the sleep state was lost and the bot would run again, wasting tokens on duplicate scans.

This is exactly what's happening in the quality monitor instance, so we've proven the case that the behavior needs to be corrected.

bot/run.py
```diff
@@ -489,6 +489,20 @@ def shutdown(sig, frame):
...
                     )
                     error_sleep = min(config.interval * (2**consecutive_preflight_errors), 300)
+                    if instance_config.source == "scheduled":
+                        if consecutive_preflight_errors >= 3:
+                            logger.error(
+                                "Scheduled source — giving up after %d preflight errors", consecutive_preflight_errors
+                            )
+                            cleanup_between_cycles(SCRIPT_DIR)
+                            break
+                        retry_delay = 30
```
florkbr: @catastrophe-brandon the scheduled error path uses a fixed 30s delay, while the non-scheduled path right above computes `min(config.interval * (2**consecutive_preflight_errors), 300)`. With only 3 retries max, the practical difference is small (30s vs 30s/60s/120s), but the fixed delay means all 3 retries wait the same amount even if the issue is a slow-recovering upstream.

bot/run.py
```diff
@@ -568,6 +586,11 @@ def shutdown(sig, frame):
             else:
                 logger.warning("Cycle produced no result")
 
+            if instance_config.source == "scheduled":
+                logger.info("Scheduled source — exiting after cycle")
+                cleanup_between_cycles(SCRIPT_DIR)
+                break
```
florkbr, line 591: @catastrophe-brandon The loop exits via break in all cases — success, skip, and "gave up after 3 errors" all result in exit code 0. KEDA won't know whether the pod exited because it finished successfully or because it failed 3 times. If we ever want alerting on "scheduled bot gave up," we'd need a nonzero exit or a metric. Not blocking, but worth considering.

florkbr:
@Hyperkid123 are you aware of anyone relying on the current behavior for `scheduled`? I like the idea of relying on Keda versus the sleep but there are some tradeoffs. WDYT

Hyperkid123:
@florkbr yes, OBSINT agent is using it for watch duty and triage I think.

PR #429, opened 2026-08-12
REHOR-123: move plaintext Slack webhook URLs to Vault-backed secrets

vishsanghishetty:
[REHOR-123](https://issues.redhat.com/browse/REHOR-123)

## Summary

Slack incoming webhook URLs were being committed to app-interface as plaintext deploy-time parameters, and passed through bot env vars and error messages unsanitized. A webhook URL is a bearer credential — anyone who reads it can post to the team's Slack channel — and app-interface config is broadly readable, so this was a live exposure.

This PR removes plaintext Slack webhook handling end-to-end and replaces it with a Vault-backed secret reference, plus guardrails so it can't regress.

### 1. Onboarding generation now rejects plaintext webhooks
- [ ] `generate_app_interface.py` raises a `ValueError` if `slack_webhook_url` is present in the config at all (checked by key membership, not truthiness — so an empty/null value is rejected too)
- [ ] Only `slack_secret_name` is accepted — validated as a real DNS-1123 Kubernetes Secret name (per-label alphanumeric/hyphen boundaries, `.`-separated, 253-char cap) and rejected outright if it's a non-string or looks like a URL (`://` or `hooks.slack.com`)
- [ ] Validation errors never echo the rejected input back (so `main()`'s JSON error serialization can't leak a submitted webhook URL)
- [ ] `main()` catches `ValueError` from `generate()` and reports it through the existing `{"error": ...}` JSON path instead of an uncaught traceback

### 2. Deployment template sources the webhook from a Kubernetes Secret
- [ ] `deploy-template.yaml.j2` adds `SLACK_SECRET_NAME` (default `devbot-secrets`) and `SLACK_SECRET_KEY` (default `slack-webhook-url`) template parameters
- [ ] The bot deployment's `SLACK_WEBHOOK_URL` env var is now sourced via `secretKeyRef` (`optional: true`) instead of a literal parameter value — a team that doesn't set up a secret just gets no Slack notifications, never a plaintext fallback

### 3. Runtime hardening
- [ ] `SLACK_WEBHOOK_URL` added to `bot/config.py`'s `SECRET_ENV_VARS` — stripped from `os.environ` before the agent starts, so the Bash tool can never read it
- [ ] `bot/run.py` captures the webhook value *before* `sanitize_env()` strips it, then threads it explicitly through `_try_slack_digest()` and `idle_reminder.on_preflight_skip()` for the life of the process
- [ ] `slack_digest.py` / `idle_reminder.py` accept an explicit `webhook_url` param (falling back to `os.environ` only for standalone/CLI use) — a dedicated end-to-end test drives the real `sanitize_env()` + both call paths together to prove the composed invariant, not just each piece in isolation: once `sanitize_env()` has run, the webhook only flows through the pre-captured value, never back through the environment. A further test chains the *complete* contract (webhook exists → captured → sanitized → digest still works → the environment a subsequently-spawned process would inherit is clean) and proves the last step by literally spawning a subprocess and inspecting what it inherited, since `bot/agent.py`'s `run_cycle()` calls the Claude Agent SDK's `query()` without overriding `env=` — the SDK's subprocess inherits this process's environment by that same default OS mechanism
- [ ] `memory-server/bot_memory_server/tools/slack.py` adds `_sanitize_webhook_error()` — httpx exceptions embed the request URL in `str(e)`; this strips that out so a failed webhook POST never leaks the URL into a tool response that flows back into agent context
- [ ] `slack_notify` / `slack_send_digest` MCP tool parameters default `webhook_url` to `None` in the signature and resolve the `SLACK_WEBHOOK_URL` env fallback inside the function body instead — FastMCP serializes literal parameter defaults into the tool's JSON schema, which is sent to every MCP client (including the agent) via `tools/list`. An `os.environ.get(...)` default would bake the secret directly into agent-visible tool metadata if it were ever set in the memory server's own env. Verified empirically with a real `FastMCP` instance both before (secret present in schema) and after (absent) this change
- [ ] Webhook delivery failures are logged without `exc_info=True` — `httpx.HTTPStatusError.__str__()` embeds the request URL, so the traceback would put the webhook in memory-server's own logs on every failed delivery even though the returned tool response was already sanitized. Logs now use the same sanitized summary as the tool response. `idle_reminder.py`'s `send_reminder()` had the identical bug independently (it posts directly to the webhook via `httpx`, not through the memory-server MCP tool) — fixed the same way, with its own `_sanitize_webhook_error()` (duplicated rather than shared import, since `bot/` and `memory-server/` are separate deployables)

### 4. Secret-scanning guardrails (prevent recurrence)
- [ ] `.claude/hooks/scan-secrets.sh` — extended to block Slack webhook URLs (`services`/`workflows`/`triggers` paths) and Slack API tokens (`xox*-` and the documented `xapp-<ver>-<part>-<id>-<hash>` app-level token format) from being posted to Jira/GitHub/GitLab
- [ ] `.gitleaks.toml` — extends gitleaks' default ruleset plus a supplementary Slack webhook URL rule
- [ ] `.pre-commit-config.yaml` — added a `gitleaks` pre-commit hook
- [ ] `.github/workflows/secret-scan.yml` — new CI job running the checksum-verified gitleaks OSS binary directly (the `gitleaks-action` wrapper requires a paid license for org-owned repos; the underlying CLI is free)
- [ ] `.github/branch-protection/{required-checks,master-protection}.json` — `gitleaks` added as a **required** status check (it has no path filter, so it's eligible to be hard-required, unlike most other checks in this repo)

### 5. Docs updated to describe the Vault-backed contract only
- [ ] `presets/workflows/onboarding/CLAUDE.md` — `slack_webhook_url` marked as a retired/rejected alias, `slack_secret_name` documented as canonical
- [ ] `presets/workflows/onboarding/skills/generate-app-interface/SKILL.md` — Slack Webhook Secret section with Vault provisioning steps
- [ ] `docs/onboarding-new-instance.md` — parameter reference tables updated; removed an incorrect instruction to override `SLACK_SECRET_KEY` (it's fixed at `slack-webhook-url`, not configurable)

## Why

- Webhook URLs are unrotatable-by-default bearer tokens embedded in plaintext GitOps config — a broad-read-access leak (or just git history) exposes them permanently until manually rotated
- The bot's own agent (with Bash access) previously had `SLACK_WEBHOOK_URL` sitting in its environment for the whole process lifetime
- Nothing previously stopped a webhook URL from being echoed back through an MCP tool's error response — or its schema, or its server-side logs — into agent/Jira/PR context or the memory server's own logs

## What stays the same

- Teams that don't want Slack notifications: no config needed, no behavior change
- Existing `devbot-secrets` shared secret (GitHub/GitLab/GPG/GCP creds) — untouched, still shared across instances
- `SLACK_NOTIFY_MODE` / digest behavior — untouched, only the credential-sourcing mechanism changed

## Out of scope (tracked separately per the plan, not code changes in this repo)
- [ ] Sweep existing app-interface deploy files for any already-committed plaintext webhook URLs and replace with Vault secret refs
- [ ] Rotate any webhook URLs that were ever committed in plaintext (treat as compromised regardless of current validity)
- [ ] Link the app-interface cleanup MR(s) and rotation confirmation back to this Jira ticket

## Validation

- [x] Full test suite: 957 passed (root, 2 xfailed/1 xpassed unrelated to this PR) + 31 passed (memory-server slack tests) — 0 failures
- [x] New/updated tests: `test_sanitize_env.py`, `test_idle_reminder.py` (explicit-webhook passthrough + a log-leak regression for `send_reminder()`'s direct httpx path), `test_slack_digest_cmd.py`, `test_slack_digest_runner.py`, `test_webhook_env_invariant.py` (composed capture→sanitize→explicit-param invariant across both call paths, plus a full-contract test that spawns a real subprocess to prove the environment it inherits is clean), `memory-server/tests/test_slack_digest.py` (webhook-leak-proof error sanitization + tool-response sanitization + server-log sanitization + a real-`FastMCP`-instance regression test asserting the webhook never appears in the generated MCP tool schema), `test_generate_app_interface.py` (secret-name validation: URL-shaped, non-string, invalid DNS-1123 labels, dotted names, empty-string retired-key rejection), `test_generate_instance.py` (secretKeyRef wiring + defaults)
- [x] `ruff format --check` / `ruff check` — clean across all changed files
- [x] `mypy` — no issues
- [x] Manual regex smoke test of updated `scan-secrets.sh` patterns against constructed sample strings (workflows/services/triggers URLs, xapp/xoxb tokens, clean text) — all correctly classified
- [x] JSON/TOML/YAML/bash syntax validated for all new config files (`.gitleaks.toml`, `secret-scan.yml`, `.pre-commit-config.yaml`, branch-protection JSON)
- [x] Pre-commit hooks (including the new `gitleaks` hook) ran clean on every commit
- [x] MCP tool schema empirically verified secret-free using a real `FastMCP` instance (before/after comparison), not just a mocked capture fixture
- [ ] CI checks on this PR (pending — will confirm green before merge)


vishsanghishetty:
@Hyperkid123 ready for review

PR #432, opened 2026-08-13
fix(konflux): bump Tekton task bundle refs to latest trusted digests

petrsimon:
## Summary

- EC (Enterprise Contract) rejects builds because `prefetch-dependencies` 0.6.0 digest was explicitly revoked (`deny_rule`) from the trusted task list
- Bumps `prefetch-dependencies` from 0.6.0 to 0.9.0 across all 6 pipeline definitions
- All other task refs already match `konflux/references/master`

> **Note:** Per-component EC checks on this PR may still show failures — they evaluate images built *before* this fix. The **PR group EC** check runs against images built with the updated refs and should pass.

## Test plan

- [ ] Konflux PR pipeline builds pass (using `prefetch-dependencies:0.9.0`)
- [ ] PR group EC verify passes
- [ ] Push pipeline succeeds after merge

### DIFF
--- a/.tekton/platform-frontend-ai-dev-memory-server-pull-request.yaml
+++ b/.tekton/platform-frontend-ai-dev-memory-server-pull-request.yaml
@@ -310,7 +310,7 @@ spec:
         - name: name
           value: prefetch-dependencies
         - name: bundle
-          value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies:0.6.0@sha256:278c47d3bf6d55eaef26453182356a819cb376ba7f55f6e8db01d9e79c98b702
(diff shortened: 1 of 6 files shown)


review-notes/RHCLOUD-48014-digest-all-paths/README.md
```diff
@@ -0,0 +1,69 @@
+# PR Review Notes — #426
```
florkbr: @petrsimon did you mean to check this in? Do you mind if we go with: https://github.com/OpenShift-Fleet/rehor/pull/434?

florkbr: I updated the description to indicate why we are still seeing the EC's fail in the PR (but not the PR group EC)

petrsimon: oh no 🫤 Thanks

PR #464, opened 2026-09-02
docs: add OpenCode migration design

Hyperkid123:
### Description

Add migration design covering Claude Code to OpenCode/TypeScript, Vertex auth to direct OpenAI API keys, proxy header injection, server/client reuse, MCP, personas, preflights, plugins, rollout, security, observability, and rollback.

[REHOR-35](https://redhat.atlassian.net/browse/REHOR-35)

### Blast radius

Documentation only. Updates MkDocs navigation and documentation index. No runtime, deployment, or proxy behavior changes.

### Rollback plan

Git revert is sufficient.

### Checklist
- [x] Container images pinned to specific tags, not latest

### AI disclosure
Assisted by: Claude Code


docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
+# OpenCode Migration Design
+
+Migration plan for moving Rehor bot instances from Claude Code plus
+`claude-agent-sdk` to OpenCode plus its TypeScript SDK. This design also
```
petrsimon, line 4: @Hyperkid123 Are you completely sold on OpenCode? I tried comparing it with pi and I'm prepared to do some lobbying for it :) Either seems like a fine choice, just so that we know the pros/cons.
https://docs.google.com/document/d/1ePFGlZFmZEk_jlL_zp_wtkprE0Ocr5Ng8uuP2B94upw/edit?tab=t.0#heading=h.szuf2vvwcp6x

Hyperkid123: Mostly yes. The plugins dsk https://opencode.ai/docs/plugins/ will be important. I know pi has some proposal but opencode seems to be a bit more mature. Its almost a coin toss TBH

petrsimon: Yes, agreed, mostly a coin toss. I like pi more because it's bare bones by default and we have all control, but I already use it myself, so I'm up for getting more familiar with opencode. The eventual upgrade to v2 which is more what we want, but still private beta or something, hopefully won't be too complicated either. I'll go through this PR tomorrow.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
+# OpenCode Migration Design
```
petrsimon, line 1: I will paste bigger chunks which I have from my exploration I did before.

Summary for this part:
- Rewrite Rehor coordinator in TypeScript, running on pinned Bun release.
 - Keep Node initially where existing MCP/Playwright tooling requires it; do not make scriptc part of migration.
 - Keep Rehor workflow, policy, integrations, persistence, and telemetry above narrow AgentRuntime boundary with versioned task, event, checkpoint, and capability schemas.
 - Implement OpenCode v1 through stable server/client SDK first, pinning exact release.
 - Run one runner-owned OpenCode server per Rehor instance, reused across sequential cycles.
 - Create one OpenCode session per cycle/task and bind every operation explicitly to its workspace.
 - Never share OpenCode servers across Rehor instances. Future concurrent cycles require isolated worktrees, event ownership, cancellation, and and budgets.
 - Normalize OpenCode events into versioned Rehor events while retaining redacted raw runtime events for debugging and migration.
 - Treat OpenShell, executor/integration proxies, NetworkPolicy, and model proxy as security boundaries; OpenCode permissions remain defense in depth.
 - Keep provider selection independent from runtime selection. Vertex and OpenAI routes may coexist during rollout.
 - Keep deterministic preflights, authorization, publication, and security gates outside OpenCode plugins.
 - Preserve and test CLAUDE.md and .claude/skills compatibility during canary migration; do not assume identical v2 behavior.
 - Treat OpenCode v2 embedded hosting as later AgentRuntime implementation, not initial migration assumption.

Initial implementation: OpenCodeV1Runtime. Later: OpenCodeV2Runtime. Coordinator, events, telemetry, policy, and persistence remain unchanged.

I assume we will also keep Vertex AI and use opus for harder reviews.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+  Squid
+```
+
+Current agent entry point is `bot/agent.py`. Preflight scripts are discovered
+and executed by `bot/preflight.py`. Deployment and proxy credentials are
+defined in `deploy/template.yaml`.
+
+## Target Architecture
+
+```text
```
petrsimon: Proposed runtime shape

 ```text
   Rehor instance pod
     Bun coordinator — long-lived
       OpenCodeRuntime adapter
         OpenCode process supervisor
           one OpenCode v1 server — long-lived
         client scoped to workspace
         one new session per cycle/task
 ```

 Current file lock keeps cycles sequential. Therefore persistent server gains:

 - startup cost paid once;
 - MCP/plugin initialization reused;
 - centralized health supervision;
 - potential session recovery;
 - natural future support for specialists.

 Why adapter still matters

 Adapter is not about reconsidering OpenCode. It protects Rehor from OpenCode v1→v2 changes:

 ```text
   Today:
   OpenCodeRuntime
     → @opencode-ai/sdk
     → child opencode serve
     → HTTP/SSE

   Later:
   OpenCodeRuntime
     → published v2 embedded SDK
     → in-process Effect host
 ```

 Rehor coordinator should never expose generated OpenCode SDK types. Adapter owns:

 - session creation and cancellation;
 - workspace/directory binding;
 - raw OpenCode → versioned Rehor event mapping;
 - usage and transcript extraction;
 - server readiness, restart, and shutdown;
 - v1/v2 config translation.

 Keep adapter small. No generic multi-harness framework needed.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+bot pod
```
petrsimon, line 53: v1
 ```text

   AgentRuntime
     → OpenCodeRuntime
         → V1ServerDriver
             → @opencode-ai/sdk
             → child `opencode serve`
             → HTTP/SSE
```

v2
```
   Later:
   AgentRuntime
     → OpenCodeRuntime
         → V2EmbeddedDriver
             → published embedded SDK
             → in-process scoped host
 ```

 Coordinator consumes only Rehor types:

 ```ts
   interface AgentRuntime {
     start(): Promise<RuntimeCapabilities>
     run(input: RehorRun, signal: AbortSignal): AsyncIterable<RehorEvent>
     stop(): Promise<void>
   }
 ```

Adapter owns:

 - server lifecycle through supervisor;
 - workspace-bound client creation;
 - session creation and explicit interruption;
 - OpenCode event ordering, deduplication, and normalization;
 - redacted raw-event retention;
 - usage, transcript, and terminal-state extraction;
 - version-specific config rendering;
 - health, version, and capability validation.

 Adapter does not own:

 - workflow decisions;
 - authorization or publication policy;
 - durable task state;
 - provider/model policy;
 - integration credentials;
 - security enforcement.

 Important v1→v2 rule

 Do not leak these outside driver:

 - generated SDK request/response types;
 - singular v1 session.* API names;
 - v1 SSE event unions;
 - v1 config fields such as provider, agent, permission, and small_model;
 - OpenCode session IDs as Rehor run IDs.

 V2 changes these areas substantially. Rehor IDs and schemas must remain independent.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+  TypeScript runner
+    OpenCode server (:4096, localhost or pod-local)
+    OpenCode SDK client
```
petrsimon: Replace:

 ```text
   OpenCode server (:4096, localhost or pod-local)
   OpenCode SDK client
 ```

 With:

 ```text
   OpenCodeRuntime
     V1ServerDriver
       runner-owned OpenCode server
       bound to 127.0.0.1
       no Kubernetes Service
       one workspace-scoped client per run
 ```

 Avoid fixed port in architecture. Port is implementation config. If 4096 remains, it must stay loopback-only.

 Do not put OpenCode in separate sidecar initially. Child process gives:

 - direct lifecycle ownership;
 - one release unit;
 - simpler shutdown and readiness;
 - easier future replacement with embedded v2;
 - no unnecessary container boundary.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+      sessions: one per cycle/task
+      SSE events: lifecycle, tool, usage, error
+    AGENTS.md / compatible CLAUDE.md
+    .opencode/agents, plugins, skills
+    MCP: memory, Jira, browser
+    HTTP: OpenAI-compatible model proxy, Git proxy, Squid
+
+central proxy pod
+  OpenAI-compatible model proxy (:8450)
+    direct OpenAI API key from Vault
```
petrsimon: I would leave our options open unless you know we won't be able to use vertex anymore.

```text
   OpenCode
     → Rehor model gateway
         ├── Vertex route
         └── OpenAI route
 ```

 OpenAI endpoint can still be first new implementation. Provider-neutral architecture prevents runtime rollout from depending on provider rollout and supports independent rollback.

File: docs/migrations/opencode-migration.md [hunk 7]
```diff
@@ -0,0 +1,484 @@
...
+  Jira MCP proxy
+  Git auth proxy
+  Executor server
+  Squid
+```
+
+OpenCode is an agent runtime, not replacement for every existing proxy. The
+proxy remains credential and egress boundary. OpenCode remains execution
+boundary inside each bot pod.
+
```
petrsimon, line 81: Target omits preflights. Add them above AgentRuntime:

 ```text
   schedule
     → deterministic preflight
     → policy decision
     → AgentRuntime.run(...)
     → normalized events
     → persistence/publication
 ```

 Do not migrate these into OpenCode plugins:

 - ticket selection;
 - merge/publication authorization;
 - model allowlisting;
 - workspace assignment;
 - credential decisions;
 - required/advisory preflight semantics.

 Plugins may provide context, telemetry, custom tools, and non-authoritative lifecycle integration.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+const opencode = await createOpencode({ hostname: "127.0.0.1", port: 4096 })
+const client = opencode.client
+```
+
+For an already-running server:
+
+```ts
+import { createOpencodeClient } from "@opencode-ai/sdk"
+
+const client = createOpencodeClient({
```
petrsimon, line 99: This is useful for prototypes and tests, but it is not a production process supervisor. Its current implementation:

- calls `createOpencodeServer()` and then creates a client using only the returned `baseUrl`;
- launches `opencode serve` through `cross-spawn`;
- resolves `opencode` from `PATH` rather than requiring an exact configured binary path;
- copies the coordinator's entire `process.env` into OpenCode;
- adds the complete config through `OPENCODE_CONFIG_CONTENT`;
- waits for stdout text beginning with `opencode server listening` and parses the URL from that human-oriented line;
- uses a five-second startup timeout by default;
- captures startup output in one unstructured string;
- exposes a server handle containing only `url` and `close()`;
- does not expose the child PID, post-readiness exit monitoring, restart policy, health state, or structured stdout/stderr events;
- does not check `/global/health` or verify server version against the SDK version;
- does not calculate or expose the active configuration hash;
- creates its client without an explicit repository directory;
- does not define behavior for config changes or a server crash during an active run.

Rehor should launch `opencode serve` through its own supervisor and use the stable client-only SDK for protocol access:

```ts
const server = await supervisor.start({
  binary: runtimeManifest.opencodeBinary,
  hostname: "127.0.0.1",
  port: runtimeConfig.port,
  environment: buildOpenCodeEnvironment(runtimeConfig),
  config: renderOpenCodeV1Config(instanceConfig),
  startupTimeoutMs: runtimeConfig.startupTimeoutMs,
})

const client = createOpencodeClient({
  baseUrl: server.url,
  directory: run.worktree,
})
```

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+  baseUrl: "http://127.0.0.1:4096",
+})
+```
+
+One server can serve multiple clients. Each client should use a separate
```
petrsimon: One server per instance improves startup cost but creates state that must be bounded:

- **Stale configuration:** hash effective runtime config and restart between runs when it changes; do not depend on undocumented hot reload.
- **Session/database growth:** establish session, event, log, snapshot, and managed-tool-output retention with quotas and cleanup metrics.
- **Workspace cross-talk:** always create directory-scoped clients, attach workspace and repository-snapshot identity to events, and dispose
project state when a worktree is removed.
- **Plugin drift:** OpenCode automatically installs configured npm plugins with Bun at startup. Production images must pre-bake and pin revie
wed plugin dependencies or use a controlled internal source; do not permit unpinned runtime installation through intercepted egress.
- **Language-server downloads:** set `"lsp": false` explicitly and `OPENCODE_DISABLE_LSP_DOWNLOAD=true` as defense in depth for the initial r
untime. Verify zero LSP download attempts and no residual language-server processes.
- **Failure concentration:** a server crash interrupts the active run and invalidates all clients. Persist partial events before restart and
never infer successful continuation.
- **Local API exposure:** model-controlled shell processes can reach loopback. Test direct API calls, session enumeration, extra model-call c
reation, permission changes, and budget bypass attempts.
- **Resource leakage:** verify child sessions, MCP transports, browser processes, shell children, and SSE readers terminate on run cancellati
on and server replacement.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+session. This is suitable for multiple logical cycles in one bot pod.
+
+Shared-server restrictions:
+
+- Server project and working-directory context are shared.
```
petrsimon: One server can host several directory-scoped clients and sessions. Every Rehor client must provide an explicit repository worktree director
y. Process-global caches, plugin loading, runtime storage, and server lifecycle remain shared across the Rehor instance.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+- Sessions and event streams need client-side ownership checks.
+- Concurrent sessions must not modify the same worktree.
+- OpenCode server Basic Auth is coarse-grained, not task authorization.
+- A server exposes powerful filesystem, shell, MCP, and session APIs.
+
```
petrsimon: One server per instance improves startup cost but creates state that must be bounded:

- **Stale configuration:** hash effective runtime config and restart between runs when it changes; do not depend on undocumented hot reload.
- **Session/database growth:** establish session, event, log, snapshot, and managed-tool-output retention with quotas and cleanup metrics.
- **Workspace cross-talk:** always create directory-scoped clients, attach workspace and repository-snapshot identity to events, and dispose
project state when a worktree is removed.
- **Plugin drift:** OpenCode automatically installs configured npm plugins with Bun at startup. Production images must pre-bake and pin revie
wed plugin dependencies or use a controlled internal source; do not permit unpinned runtime installation through intercepted egress.
- **Language-server downloads:** set `"lsp": false` explicitly and `OPENCODE_DISABLE_LSP_DOWNLOAD=true` as defense in depth for the initial r
untime. Verify zero LSP download attempts and no residual language-server processes.
- **Failure concentration:** a server crash interrupts the active run and invalidates all clients. Persist partial events before restart and
never infer successful continuation.
- **Local API exposure:** model-controlled shell processes can reach loopback. Test direct API calls, session enumeration, extra model-call c
reation, permission changes, and budget bypass attempts.
- **Resource leakage:** verify child sessions, MCP transports, browser processes, shell children, and SSE readers terminate on run cancellati
on and server replacement.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+  no Node installation
+```
+
+The safest target is a native Rehor runner that connects to an already-running
+OpenCode server through HTTP. This avoids compiling OpenCode internals and
+keeps plugin loading under its supported runtime. An alternative is compiling
+small standalone TypeScript preflight utilities, provided they remain free of
+unsupported npm/runtime APIs.
+
+### Acceptance Test Before Adoption
```
petrsimon, line 144: Few notes for bun

Before making Bun the production coordinator runtime, verify:

- exact Bun version and artifact checksum are pinned for Linux amd64 and arm64;
- installation works in the UBI/Konflux build path without an unpinned network fetch;
- generated SBOM and vulnerability scanning include Bun, OpenCode, and JavaScript dependencies;
- organization-specific FIPS/TLS requirements are satisfied or an explicit exception exists;
- UBI glibc, CA trust, proxy variables, and certificate behavior work for all coordinator HTTP/SSE traffic;
- native dependencies and optional postinstall scripts are either removed, prebuilt, or reproducible;
- `bun install --frozen-lockfile` performs no unexpected runtime installation;
- OpenCode plugins and tools are pre-baked and pinned rather than fetched dynamically at startup;
- the runtime works with `readOnlyRootFilesystem`, with every writable path redirected to reviewed mounts;
- no application dependency relies accidentally on global npm state from the existing image.

### Compatibility tests

Initial Bun acceptance suite:

1. Start the Bun coordinator under `tini`.
2. Launch the exact OpenCode binary through the Rehor supervisor.
3. Validate `/global/health`, version match, and config hash.
4. Create a worktree-scoped SDK client and complete a session.
5. Consume long-lived SSE without an implicit request timeout.
6. Abort an active session through `AbortSignal` and verify terminal event handling.
7. Terminate the coordinator and verify OpenCode, shell, MCP, browser, and SSE child resources are reaped.
8. Kill OpenCode during a run and verify partial events/usage persist and the run becomes interrupted.
9. Restart OpenCode between runs and verify stale sessions/config do not leak into the next run.
10. Run existing Node-based MCP and Playwright paths alongside Bun.
11. Run existing Python preflights unchanged.
12. Verify proxy, CA, Git/executor, filesystem, and read-only-root behavior on amd64 and arm64.
13. Compare startup time, RSS, CPU, and image size against the current runtime, but do not block migration on improvements unless thresholds are declared beforehand.


### V2 upgrade implication

The Bun decision should simplify, not prematurely implement, v2:

```text
initial
  Bun coordinator
    → OpenCodeV1ServerDriver
      → separate Bun-compiled OpenCode binary

future
  same Bun coordinator boundary
    → OpenCodeV2EmbeddedDriver
      → published in-process OpenCode host
```

Before adopting embedded v2, re-run the compatibility suite with the exact Bun/OpenCode package combination and add tests for scoped resource disposal, database ownership, durable session-event cursors, ephemeral live-event loss, plugin/config migration, and process-level failure isolation lost by moving in-process.

docs/migrations/opencode-migration.md
```diff
@@ -0,0 +1,484 @@
...
+- SSE event stream handles long responses and disconnects.
+- MCP configuration and tool calls remain functional.
+- Signals, timeouts, child processes, Git, and filesystem behavior match current runner.
+- Native and normal-runtime runs produce equivalent cycle results and cost data.
+
+Until this test passes, use a supported OpenCode distribution and treat scriptc
+as an optimization experiment. Do not make native compilation a dependency of
+initial OpenCode migration.
+
+## Direct OpenAI Authentication
```
petrsimon, line 159: The PR currently standardizes on `/v1/chat/completions` through `@ai-sdk/openai-compatible`. Current OpenCode documentation and source distinguish two paths:

- `@ai-sdk/openai-compatible` for Chat Completions-compatible `/v1/chat/completions` endpoints;
- `@ai-sdk/openai` for OpenAI's `/v1/responses` endpoint.

The inspected OpenCode provider implementation selects `sdk.responses(modelID)` for the native `openai` provider. Newer OpenAI reasoning and tool behavior may depend on Responses API semantics. Therefore:

- select exact models first;
- test their required endpoint, streaming, reasoning, tool-call, structured-output, and usage behavior;
- use native `@ai-sdk/openai` when the selected model uses Responses API;
- use `@ai-sdk/openai-compatible` only for a verified Chat Completions-compatible route;
- do not translate Responses requests into Chat Completions inside a header-injection proxy;
- support both routes during canary only if real model requirements justify both.

The gateway is protocol pass-through plus authentication/policy. Any request/response schema translation must be a separately named and tested adapter.

### Credential and caller-authentication boundaries

Required flow:

```text
OpenCode provider client
  → authenticated Rehor model gateway
      → validate caller, route, model, limits, and policy
      → discard incoming provider authorization
      → inject proxy-owned OpenAI credential
      → fixed OpenAI upstream
```

docs/migrations/opencode-migration.md
Hyperkid123: Yes, the first migration round should not affect any tenants or models. Its just a prep for the provider switch.

I think this sums it up nicely.

docs/migrations/opencode-migration.md
Hyperkid123: We need to support both, so yeah we should be able to handle both. Need to update this.

docs/migrations/opencode-migration.md
Hyperkid123: It was just a prototype to show thecase interafce, no production code examples are in this doc.

docs/migrations/opencode-migration.md
Hyperkid123: The TLDR is we will not share the server, maybe something to think about later on if we do some persistent volumes, but opencode server sharing is not possible with current approach.

docs/migrations/opencode-migration.md
Hyperkid123: Srvers can be started with each cycle. I don't think we want to keep t alive. Especially with some of the future improvements I have in mind.

Hyperkid123:
@petrsimon the doc should be updated

petrsimon (approved):
Let's chop it up into tickets :)

PR #474, opened 2026-09-07
feat(runtime): define and harden coordinator contract

petrsimon:
## Summary

- define versioned coordinator run and self-describing event contracts with schema-backed runtime validation
- add deterministic event-ledger rules for sequence gaps, replay, attribution, usage, and terminal state
- add a run-bound event factory so adapters stamp correct envelope identity while preserving per-event model drift
- remove duplicated usage payload `attemptId`/`provider` fields; envelope remains authoritative
- publish coordinator entry point, CI checks, dependency audit coverage, and architecture/migration docs
- keep production execution on `bot/run.py` and `bot/agent.py`; runtime adapter remains future work

## Validation

- coordinator frozen install, 25 Bun tests, TypeScript check, build, and `bun audit` passed
- `make verify` — Python, Go, dashboard, and coordinator checks passed
- `make memory-verify` with throwaway pgvector PostgreSQL — 228 tests passed; existing dependency advisories remain report-only
- strict MkDocs build and repository format/schema checks passed
- GitHub `e2e-minimal`, `coordinator`, and `coordinator-audit` checks passed
- local container verification was initially blocked by Podman VM storage while committing the existing CUDA/PyTorch memory-server layer; GitHub Linux E2E passed after the entrypoint check fix

---

### Description

Establishes a validated coordinator boundary before adapter implementation. Events remain self-describing persistence units; adapters use the event factory to avoid hand-copying envelope fields. No downstream runtime changes yet; existing Python execution remains authoritative.

https://issues.redhat.com/browse/REHOR-138

---

### Blast radius

Coordinator package, repository CI, and architecture documentation. Compatibility was checked against existing bot and memory-server tests; no production code imports the coordinator package yet.

---

### Rollback plan

Git revert is sufficient. Coordinator remains scaffolding and does not alter production execution.

---

### Checklist
- [ ] Tested against at least one consuming repo/service — no coordinator runtime consumer exists yet
- [x] Container images pinned to specific tags, not `latest` — no container image references changed

### AI disclosure

Assisted by: OpenAI Codex (review, implementation, tests, and documentation).


coordinator/bun.lock
Hyperkid123, line 1: Do we really need bun? We should be able to run this in node. What is the RH support for bun runtime? If we do not have a proper UBI for bun we will have to use node instead. I know Bun performance gains etc, but the base image availability will be the decider here.

petrsimon: We can use node for sure. I chose bun because opencode runs on it. And yes, we would have to install bun into the images just like we do with other stuff and namely node 22 :), no bun in RH images. Let me know I switch it to node, no strong opinion here. But it could be faster at least on paper ;)

Hyperkid123: The way I am looking at is if RH registry contains a "prepared" ubi with node we do have some support. You know like ubi9/node(maybe not the exact name). We have that for go, node, python, etc. Do we have thet for bun? If not, we should probably go with node instead.

petrsimon: Yes, I checked again and there's no bun. Agent tested it in ubi image and looks ok, the but there issues which will be relevant for opencode runtime as well:
- AVX2 / SIGILL on older x86 nodes
- no grype for bun, node 22 is checked

I don't have access to nodes, but we should check by 

1. Cheap first pass — instance types
`oc get nodes -L node.kubernetes.io/instance-type`
Anything Haswell-or-newer (AWS m5/c5/r5/m6i/c6i/m7i, any 2013+ CPU) has AVX2. Only pre-2013 Sandy/Ivy Bridge lacks it.

2. Confirm on the nodes
```
for n in $(oc get nodes -o name); do
  echo -n "$n: "; oc debug "$n" -q -- chroot /host grep -om1 avx2 /proc/cpuinfo || echo MISSING
done
```
Or, if NFD runs: oc get nodes -l feature.node.kubernetes.io/cpu-cpuid.AVX2=true

Hyperkid123: Can't check either, we may just have to try and see. I do have access to terminal execution, any code snippet form node runtime that would be able to validate?

petrsimon: Of course, silly me

```
oc exec <pod> -- bash -lc '
  grep -m1 "model name" /proc/cpuinfo
  grep -qw avx2 /proc/cpuinfo && echo "AVX2: yes — OK" || echo "AVX2: NO — SIGILL risk"
'
```

I ran it and it looks ok. 

```
for p in $(oc get pods -o name); do
  oc exec "${p#pod/}" -- bash -lc \
    'grep -m1 "model name" /proc/cpuinfo; grep -qw avx2 /proc/cpuinfo && echo AVX2-OK || echo NO-AVX2' 2>/dev/null
done | sort | uniq -c
  20 AVX2-OK
  20 model name	: Intel(R) Xeon(R) Platinum 8488C
```

So unless there are some old workers in that node pool we should be fine. Should we try to verify that it can't happen and that all workers are fairly modern? It's just `staging` after all 😅 

Regarding node.js, I guess it's safer path, at least we get warning from grype about the CVEs. wdyt? Or bun?

Hyperkid123: Yeah I think we are good. Lets go with node for now then

Hyperkid123 (approved):

PR #481, opened 2026-09-08
REHOR-74: separate integration tests from unit CI collection

vishsanghishetty:
## Summary
This PR implements REHOR-74 by separating Docker-dependent integration tests from default CI/unit test collection, while restoring root `tests/` coverage that had been silently excluded.

## Problem
- Root `tests/` were not part of `pytest` `testpaths`, so unit tests there were not running in CI.
- Integration tests that require external services needed to remain runnable locally but never be collected in default CI test runs.

## What Changed
- Moved git-proxy integration tests into `tests/integration/`.
- Added integration test support files:
  - `tests/integration/__init__.py`
  - `tests/integration/conftest.py`
  - `tests/integration/README.md`
- Updated `pyproject.toml`:
  - Added `tests` to `testpaths` so root unit tests are discovered.
  - Added `--ignore=tests/integration` to default `addopts` so integration tests are excluded from default collection.
  - Registered `integration` pytest marker.
- Added Make targets:
  - `test-unit` (CI-equivalent unit run)
  - `test-integration` (explicit integration run)
- Restored `tests/test_new_work.py` to preserve intended root test coverage for REHOR-74 validation.
- Resolved merge conflicts with latest `master` in `Makefile` without dropping REHOR-74 targets.

## Verification
Re-ran checks before pushing:

### Python
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run mypy`
- `uv run pytest`

### Go
- `cd proxy/executor && go vet ./...`
- `cd proxy/executor && go test -race ./...`

### Frontend / Dashboard
- `npx -y @biomejs/biome@2.5.12 check .`
- `cd dashboard && npm run typecheck`
- `cd dashboard && npm run build`
- `cd dashboard && npm test`

Targeted REHOR-74 checks:
- `uv run pytest tests/test_merge.py tests/test_new_work.py tests/test_transcripts.py tests/test_cost_tracking_logging.py -v`
- `uv run pytest tests --collect-only -q`
- `uv run pytest tests/integration/ --collect-only -q`

## Expected Behavior After Merge
- Default CI/unit pytest runs include root `tests/` unit tests.
- Integration tests in `tests/integration/` are excluded from default collection.
- Integration tests remain runnable explicitly via path or `make test-integration`.

## Risk / Notes
- No production runtime behavior changes; scope is test layout/config and Make targets.
- Local plan artifacts are intentionally kept out of commits.


vishsanghishetty:
@Hyperkid123 ready for review

vishsanghishetty:
@Hyperkid123 ready for review

tests/integration/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+    make test-integration
+"""
+
+import pytest
+
+
+def pytest_collection_modifyitems(items):
+    """Automatically mark all tests in this directory as integration."""
+    for item in items:
+        item.add_marker(pytest.mark.integration)
```
petrsimon: This one we should fix, I think:

`pytest_collection_modifyitems` is a session-scoped hook. Even when declared in a subdirectory `conftest.py`, the `items` it receives is the *full* collected list for the run, not the subset under `tests/integration/`. So the docstring's "all tests in this directory" isn't what happens.

Reproduced on this branch:

```
$ uv run pytest tests/integration tests/test_merge.py --collect-only -q -m integration
38 tests collected
```

Nine of those are the real integration tests; the other 29 are `tests/test_merge.py`, marked `integration` purely by being in the same session. Any invocation that mixes an integration path with a unit path mismarks everything, which makes `-m integration` / `-m "not integration"` unusable as a selector - and registering the marker in `pyproject.toml` implies that is the intended use.

The default CI run is safe today only because `--ignore=tests/integration` stops this conftest from ever loading. That's an accident, not a guarantee.

Filter by path:

```python
from pathlib import Path

def pytest_collection_modifyitems(items):
    here = Path(__file__).parent
    for item in items:
        if here in Path(str(item.path)).parents:
            item.add_marker(pytest.mark.integration)
```

Or drop the hook entirely and put `pytestmark = pytest.mark.integration` at the top of each module in this directory - less magic, and it can't leak.

tests/test_cost_tracking_logging.py
```diff
@@ -108,7 +108,7 @@ def test_cost_push_http_error_now_detected(tmp_costs_file, mock_result, caplog):
 # --- Test 2: Status push failures are completely silent ---
 
 
-@pytest.mark.asyncio
+@pytest.mark.anyio
```
petrsimon, line 111: Please add an explicit `anyio` to the dev extras, or `pytest-asyncio` plus `asyncio_mode` if you'd rather keep the original marker.

it only works because the `anyio` pytest plugin arrives transitively through `httpx`. Neither `anyio` nor `pytest-asyncio` appears in `[project.optional-dependencies].dev`.

`addopts` includes `--strict-markers`. If that transitive dependency ever moves, or a run uses `-p no:anyio`, the `anyio` marker is unregistered and pytest errors during collection - the *entire* session fails, not just this test, and it fails pointing at a marker rather than at a missing dependency.

Makefile
```diff
@@ -125,6 +125,12 @@ container-verify: ## Run container build + smoke checks locally (CI-equivalent,
 	@echo ""
 	@echo "All container verification checks passed."
 
+test-unit: ## Run unit tests only (same as CI)
+	uv run pytest
+
+test-integration: ## Run integration tests (requires docker-compose stack running)
+	uv run pytest tests/integration/ -v -p no:cacheprovider
```
petrsimon: "same as CI" isn't accurate. CI runs (`.github/workflows/ruff-format.yml:105`):

```
uv run pytest --cov=bot --cov-report=term-missing --cov-fail-under=50
```

This target runs bare `uv run pytest`, so it skips the coverage gate - someone using it as their pre-push check can pass locally and fail CI. Either match the flags or reword the comment. For what it's worth the gate does still pass on this branch: I measured 71.16% with the new `testpaths`.

Also `uv run pytest` is already the Python test step in `verify` (`Makefile:25`), so this duplicates it. Not a problem, but `verify` could call `test-unit` rather than repeat the command.

pyproject.toml
```diff
@@ -62,7 +63,10 @@ pythonpath = ["dashboard/tests"]
 python_files = ["test_*.py"]
 python_classes = ["Test*"]
 python_functions = ["test_*"]
-addopts = "-v --strict-markers --strict-config --tb=short --import-mode=importlib"
+addopts = "-v --strict-markers --strict-config --tb=short --import-mode=importlib --ignore=tests/integration"
+markers = [
+    "integration: requires external services (docker-compose stack, databases, etc.)",
+]
```
petrsimon: Two small notes on the exclusion mechanism.

**`--ignore` resolves relative to the invocation directory, not to `rootdir`.** So this only excludes integration tests when pytest runs from the repo root. That holds for CI and for both Make targets today, so nothing is broken - but it's a silent-wrong-answer failure mode rather than an error if someone runs pytest from a subdirectory. `norecursedirs`, or `collect_ignore` in a root `conftest.py`, is rootdir-relative and wouldn't have that property.

**Migration-facing: the boundary is declared twice.** The `integration` marker registered just below is currently decorative - nothing deselects on it, because `--ignore` has already removed the files from collection. So the split lives in `addopts` *and* implicitly in the Makefile targets, in a form specific to pytest.

Once the repo runs both pytest and a Bun test runner, that rule wants one home, and the Make targets are the natural one since they're the only layer both runners share. Not something to change in this PR, but I'd rather the marker either became load-bearing (`-m "not integration"`) or the duplication was acknowledged, before this becomes the migration's test contract.

tests/test_new_work.py
```diff
@@ -1,11 +1,13 @@
-"""Tests for new_work.py repo label matching with org/repo support."""
+"""Tests for preflight repo label matching with org/repo support."""
 
 import sys
 from pathlib import Path
 
-sys.path.insert(0, str(Path(__file__).resolve().parent.parent / ".claude" / "skills" / "new-work"))
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "presets" / "shared" / "preflight"))
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / ".claude" / "skills"))
```
petrsimon: These two `sys.path.insert(0, ...)` calls run at module import time and mutate `sys.path` for the rest of the pytest session, not just this module.

Both directories they inject hold generically-named modules - `presets/shared/preflight` has `common.py`, and `.claude/skills` has `paths.py` and its own `conftest.py`. Combined with `--import-mode=importlib` that sets up order-dependent shadowing: whichever test module imports first wins, and a failure would surface in some unrelated test rather than here.

I did check that the `.claude/skills` entry is load-bearing (`jira_sprint_preflight` imports `jira_mcp` from there), so it can't just be dropped.

Prefer scoping it:

```python
@pytest.fixture(autouse=True)
def _preflight_path(monkeypatch):
    root = Path(__file__).resolve().parent.parent
    monkeypatch.syspath_prepend(root / ".claude" / "skills")
    monkeypatch.syspath_prepend(root / "presets" / "shared" / "preflight")
```

with the imports moved inside the tests, so `monkeypatch` unwinds it afterwards.

tests/test_new_work.py
```diff
@@ -1,11 +1,13 @@
...
 
-from new_work import build_repo_lookup, match_repo_labels
+from common import build_repo_lookup
+from jira_sprint_preflight import _match_repo_labels
```
petrsimon: Minor: the file now tests a private symbol, and importing `jira_sprint_preflight` to reach it pulls in `jira_mcp` and its MCP dependencies at import time - a heavy import for what is pure string matching over a dict.

`build_repo_lookup` already lives in `common.py` and is public. Moving `_match_repo_labels` there as `match_repo_labels` would let this file import one module instead of two, drop the `.claude/skills` path insert entirely, and stop the test depending on a private name the next refactor is free to rename.

tests/integration/README.md
```diff
@@ -0,0 +1,89 @@
+# Integration Tests
```
petrsimon, line 1: Two small things to improve:

- `tests/README_GIT_PROXY_TESTS.md` still exists at the `tests/` root and is now stale: it describes this same suite at its old path. Since this README supersedes it, please delete it in this PR so there's one source of truth.
- Worth documenting here that these tests **skip** rather than fail when the stack is down. Right now `pytest.skip("Proxy service not healthy - is docker-compose running?")` means a run with no Docker at all reports green, which is correct on a laptop and wrong for the scheduled job I'd like to see running this (see the migration notes in the review body). An env gate - `REHOR_INTEGRATION_REQUIRED=1` turning those skips into failures - would let one suite serve both.

petrsimon (approved):
thanks a lot, few small issues we found with my agent :)

petrsimon (changes requested):
sorry, misclick some changes would be nice to have :)

petrsimon (approved):

PR #493, opened 2026-09-09
REHOR-125 - support workflow model defaults and instance overrides

vishsanghishetty:
### Description
[REHOR-125](https://issues.redhat.com/browse/REHOR-125)

Allow workflow presets to declare their own default model (`default_model` in `manifest.yaml`), while supporting per-instance overrides (`model` in `instance.yaml` or `BOT_MODEL` env var) and falling back gracefully to the global model defined in `config.json`.

Resolves model selection following a strict 4-tier precedence hierarchy:
1. `instance.yaml` `model` (explicit pin)
2. `BOT_MODEL` environment variable (deploy overlay)
3. workflow `manifest.yaml` `default_model`
4. global `config.json` `claude.model` (fallback)

#### Changes Made:
- **`bot/config.py`**:
  - Added `_nonempty_model` sanitization helper (strips whitespace, treats empty/non-string values as `None`).
  - Added `model: str | None = None` to `InstanceConfig` dataclass and populated it in `from_yaml()` and `from_env()`.
  - Added `resolve_cycle_model(script_dir, instance_config, global_config, remote_agent_dir=None)` implementing 4-tier resolution with source logging.
- **`bot/agent.py`**:
  - Added `model: str | None = None` parameter to `run_cycle()`, passing `model or config.model` to `ClaudeAgentOptions`.
- **`bot/run.py`**:
  - Invokes `resolve_cycle_model()` on the preflight `start` path and passes the resolved `cycle_model` into `run_cycle()`.
- **Tests & Docs**:
  - Added unit tests in `bot/tests/test_instance_config.py` covering `InstanceConfig` model loading and all precedence tiers in `resolve_cycle_model()`.
  - Added integration test in `bot/tests/test_preflight_status.py` verifying resolved model passed to `run_cycle`.
  - Updated documentation in `docs/presets/instance-config.md`, `docs/presets/custom-workflows.md`, `docs/onboarding-new-instance.md`, and `docs/migrations/preset-migration-guide.md`.

---

### Blast radius
- All bot instances consuming `instance.yaml`, `BOT_MODEL`, or preset `manifest.yaml`.
- Fully backwards compatible: when `model` and `default_model` are omitted, falls back to `config.json` `claude.model` exactly as before.
- Unit and integration tests in `bot/tests` verified passing.

---

### Rollback plan
- Standard `git revert` of this PR is sufficient.

---

### Checklist

### DIFF
--- a/bot/config.py
+++ b/bot/config.py
@@ -26,6 +26,14 @@ class Config:
     idle_reminder_cooldown_seconds: int = _DEFAULT_COOLDOWN_SECONDS
 
 
+def _nonempty_model(value: object) -> str | None:
+    """Return stripped string if non-empty, otherwise None."""
+    if not isinstance(value, str):
+        return None
+    stripped = value.strip()
+    return stripped or None
+
+
 @dataclass
 class InstanceConfig:
     """Per-instance preset selection from instance.yaml or env var fallback."""
@@ -35,6 +43,7 @@ class InstanceConfig:
     envs: list[str] | None = None  # None = all available, [] = none
     claude_md_strategy: str = "ignore"  # replace / append / ignore
     idle_cycle_limit: int = 0  # 0 = feature disabled
+    model: str | None = None
 
     @classmethod
     def from_yaml(cls, path: Path) -> InstanceConfig:
@@ -48,6 +57,7 @@ class InstanceConfig:
             envs=data.get("envs"),
             claude_md_strategy=strategy,
             idle_cycle_limit=int(data.get("idle_cycle_limit", 0)),
+            model=_nonempty_model(data.get("model")),
         )
 
     @classmethod
@@ -57,7 +67,8 @@ class InstanceConfig:
         envs: list[str] | None = None
         if envs_str is not None:
             envs = [e.strip() for e in envs_str.split(",") if e.strip()]
-        return cls(workflow=workflow, envs=envs)
+        model = _nonempty_model(os.environ.get("BOT_MODEL"))
+        return cls(workflow=workflow, envs=envs, model=model)
 
 
 def load_instance_config(remote_agent_dir: Path | None) -> InstanceConfig:
@@ -68,15 +79,21 @@ def load_instance_config(remote_agent_dir: Path | None) -> InstanceConfig:
         if yaml_path.is_file():
             ic = InstanceConfig.from_yaml(yaml_path)
             logger.info(
-                "Loaded instance.yaml: workflow=%s, source=%s, envs=%s",
+                "Loaded instance.yaml: workflow=%s, source=%s, envs=%s, model=%s",
                 ic.workflow,
                 ic.source,
                 ic.envs,
+                ic.model,
             )
             return ic
 
     ic = InstanceConfig.from_env()
-    logger.info("No instance.yaml — env/defaults: workflow=%s, envs=%s", ic.workflow, ic.envs)
+    logger.info(
+        "No instance.yaml — env/defaults: workflow=%s, envs=%s, model=%s",
+        ic.workflow,
+        ic.envs,
+        ic.model,
+    )
     return ic
 
 
@@ -217,6 +234,34 @@ def load_manifest(
         return None
     with open(path) as f:
         return yaml.safe_load(f)
+
+
+def resolve_cycle_model(
+    script_dir: Path,
+    instance_config: InstanceConfig,
+    global_config: Config,
+    remote_agent_dir: Path | None = None,
+) -> str:
+    """Resolve which model to use for the agent cycle following 4-tier precedence:
+
+    1. instance.yaml `model` (explicit pin)
+    2. BOT_MODEL environment variable (deploy overlay)
+    3. workflow manifest.yaml `default_model`
+    4. global config.json `claude.model` (fallback)
+    """
+    logger = logging.getLogger(__name__)
+    if pinned := _nonempty_model(instance_config.model):
+        logger.info("Resolved cycle model: %s (source=instance.yaml)", pinned)
+        return pinned
+    if pinned := _nonempty_model(os.environ.get("BOT_MODEL")):
+        logger.info("Resolved cycle model: %s (source=BOT_MODEL)", pinned)
+        return pinned
+    manifest = load_manifest(script_dir, instance_config.workflow, remote_agent_dir) or {}
+    if pinned := _nonempty_model(manifest.get("default_model")):
+        logger.info("Resolved cycle model: %s (source=workflow:%s)", pinned, instance_config.workflow)
+        return pinned
+    logger.info("Resolved cycle model: %s (source=config.json)", global_config.model)
+    return global_config.model
 
 
 def validate_manifest(
--- a/bot/config.py
+++ b/bot/config.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 
 import yaml
@@ -24,6 +24,7 @@ class Config:
     cycle_timeout: int
     board_key: str
     idle_reminder_cooldown_seconds: int = _DEFAULT_COOLDOWN_SECONDS
+    model_tiers: dict[str, str] = field(default_factory=dict)
 
 
 def _nonempty_model(value: object) -> str | None:
@@ -32,6 +33,24 @@ def _nonempty_model(value: object) -> str | None:
         return None
     stripped = value.strip()
     return stripped or None
+
+
+def _load_model_tiers(raw_tiers: object) -> dict[str, str]:
+    """Parse claude.modelTiers dict, dropping empty or non-string entries with warnings."""
+    logger = logging.getLogger(__name__)
+    if not isinstance(raw_tiers, dict):
+        return {}
+    tiers: dict[str, str] = {}
+    for tier, model in raw_tiers.items():
+        if not isinstance(tier, str) or not tier.strip():
+            continue
+        cleaned_tier = tier.strip()
+        cleaned_model = _nonempty_model(model)
+        if cleaned_model:
+            tiers[cleaned_tier] = cleaned_model
+        else:
+            logger.warning("Dropping invalid or empty model for tier '%s' in claude.modelTiers", tier)
+    return tiers
 
 
 @dataclass
@@ -78,6 +97,8 @@ def load_instance_config(remote_agent_dir: Path | None) -> InstanceConfig:
         yaml_path = remote_agent_dir / "instance.yaml"
         if yaml_path.is_file():
             ic = InstanceConfig.from_yaml(yaml_path)
+            if ic.model is None and (env_model := _nonempty_model(os.environ.get("BOT_MODEL"))):
+                ic.model = env_model
             logger.info(
                 "Loaded instance.yaml: workflow=%s, source=%s, envs=%s, model=%s",
                 ic.workflow,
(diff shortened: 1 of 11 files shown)


vishsanghishetty:
@Hyperkid123 ready for review

bot/tests/test_preflight_status.py
```diff
@@ -1,8 +1,12 @@
 """Preflight skip/error POST bot status so the dashboard check-in time is fresh."""
 
+import datetime
 import sys
 from unittest.mock import MagicMock, patch
 
+if not hasattr(datetime, "UTC"):
+    datetime.UTC = datetime.timezone.utc  # noqa: UP017
```
petrsimon: This shim is not needed: `pyproject.toml` requires Python `>=3.11` (`datetime.UTC` exists since 3.11) and the image runs 3.12. It only masks running the tests on 3.10 locally. Please drop it, together with the `import datetime` above.

bot/tests/test_preflight_status.py
```diff
@@ -18,6 +22,10 @@ def _mock_sdk():
     for mod_name in list(sys.modules):
         if mod_name.startswith("bot.agent") or mod_name == "bot.run":
             sys.modules.pop(mod_name, None)
+    if "bot" in sys.modules:
+        for attr in ("run", "agent"):
+            if hasattr(sys.modules["bot"], attr):
+                delattr(sys.modules["bot"], attr)
```
petrsimon: I could not find a reason for this (and the matching block in the teardown below). I removed both and ran this file and the whole suite on 3.12; everything passes, including the new `test_start_calls_run_cycle_with_resolved_model`. If it was needed on your machine, please say why in the PR; otherwise let's drop it to keep the fixture minimal.

bot/config.py
```diff
@@ -219,6 +236,34 @@ def load_manifest(
...
+    1. instance.yaml `model` (explicit pin)
+    2. BOT_MODEL environment variable (deploy overlay)
+    3. workflow manifest.yaml `default_model`
+    4. global config.json `claude.model` (fallback)
+    """
+    logger = logging.getLogger(__name__)
+    if pinned := _nonempty_model(instance_config.model):
+        logger.info("Resolved cycle model: %s (source=instance.yaml)", pinned)
+        return pinned
+    if pinned := _nonempty_model(os.environ.get("BOT_MODEL")):
```
petrsimon: `BOT_MODEL` is now read in two places: `InstanceConfig.from_env()` already sets `model` from it, and this line reads it again to cover "instance.yaml exists but omits `model`". Simpler to apply the overlay once in `load_instance_config` (after `from_yaml`, if `ic.model is None`), so `resolve_cycle_model` never touches the environment and the precedence lives in one place.

Related question on the precedence itself: the docs call `BOT_MODEL` the "deploy overlay", but it loses to `instance.yaml`. That makes it useless as an emergency lever (e.g. forcing every instance off a broken model from the deploy template without a config-repo commit). Is `instance.yaml` > env deliberate? Either order is fine with me as long as it is a conscious choice and the docs say why.

docs/migrations/preset-migration-guide.md
```diff
@@ -155,6 +156,7 @@ Instances without a config repo (or without `instance.yaml`) can configure prese
 |---------|---------|-------------|
 | `BOT_WORKFLOW_PRESET` | `jira-sprint` | Workflow preset name |
 | `BOT_ENV_PRESETS` | _(all available)_ | Comma-separated env preset names. Empty string = none. |
+| `BOT_MODEL` | _(from config.json / workflow default)_ | Model override. Unlike workflow/env presets (which are only checked when `instance.yaml` is missing), `BOT_MODEL` also overlays if `instance.yaml` exists but omits `model`. |
 
 These are checked only when no `instance.yaml` is found. If `instance.yaml` exists, it takes precedence.
```
petrsimon: This sentence now contradicts the `BOT_MODEL` row directly above it. Suggest: "`BOT_WORKFLOW_PRESET` and `BOT_ENV_PRESETS` are checked only when no `instance.yaml` is found. `BOT_MODEL` also applies when `instance.yaml` exists but omits `model`." (or adjust once the precedence question in `bot/config.py` is settled).

docs/presets/custom-workflows.md
```diff
@@ -70,6 +70,9 @@ name: my-workflow
 type: workflow
 description: Brief description of what this workflow does
 
+# Optional default model for cycles using this workflow
+default_model: claude-sonnet-4-6
```
petrsimon: Design question (not blocking this PR, but let's settle it before any built-in preset gets a `default_model`).

A concrete model ID in a shared workflow preset couples the preset to one provider. Presets are otherwise provider-neutral (instructions, skills, preflights), while the model ID is a deployment concern: it has to match the proxy allowlist (`VERTEX_ALLOWED_MODELS`) and, under the OpenCode runtime (REHOR-137), the default becomes an OpenAI-family ID resolved by a different provider. A preset that says `claude-sonnet-4-6` would then be wrong or rejected there, and every manifest would need editing on a provider change.

Proposal: let the workflow declare *what it needs*, not *which model*:

```yaml
# manifest.yaml
model_tier: light     # light | standard | heavy (or role names: review | implement)
```

and keep the tier-to-ID mapping in deployment-owned config next to the existing default:

```json
"claude": {
  "model": "claude-opus-4-6",
  "tiers": { "light": "claude-sonnet-4-6", "heavy": "claude-opus-4-6" }
}
```

Resolution stays a strict precedence, just with the workflow step going through the map: `instance.yaml model` (concrete pin, deploy knows its allowlist) > `BOT_MODEL` > `tiers[manifest.model_tier]` > `claude.model`. Unknown tier fails at startup instead of a 403 mid-cycle, which also gives us the fail-fast this PR currently lacks. It also lines up with the planned preflight complexity routing (REHOR-126), which would emit a tier rather than an ID, and with subagent tiers (REHOR-128).

`manifest.yaml` is the right file for it (better than the per-workflow `config.json` in the Jira description); only the value type should change from ID to tier.

bot/tests/test_preflight_status.py
petrsimon: This shim is not needed: `pyproject.toml` requires Python `>=3.11` (`datetime.UTC` exists since 3.11) and the image runs 3.12. It only masks running the tests on 3.10 locally. Please drop it, together with the `import datetime` above.

bot/tests/test_preflight_status.py
petrsimon: I could not find a reason for this (and the matching block in the teardown below). I removed both and ran this file and the whole suite on 3.12; everything passes, including the new `test_start_calls_run_cycle_with_resolved_model`. If it was needed on your machine, please say why in the PR; otherwise let's drop it to keep the fixture minimal.

bot/config.py
petrsimon: `BOT_MODEL` is now read in two places: `InstanceConfig.from_env()` already sets `model` from it, and this line reads it again to cover "instance.yaml exists but omits `model`". Simpler to apply the overlay once in `load_instance_config` (after `from_yaml`, if `ic.model is None`), so `resolve_cycle_model` never touches the environment and the precedence lives in one place.

Related question on the precedence itself: the docs call `BOT_MODEL` the "deploy overlay", but it loses to `instance.yaml`. That makes it useless as an emergency lever (e.g. forcing every instance off a broken model from the deploy template without a config-repo commit). Is `instance.yaml` > env deliberate? Either order is fine with me as long as it is a conscious choice and the docs say why.

bot/config.py
```diff
@@ -219,6 +236,34 @@ def load_manifest(
...
+        logger.info("Resolved cycle model: %s (source=BOT_MODEL)", pinned)
+        return pinned
+    manifest = load_manifest(script_dir, instance_config.workflow, remote_agent_dir) or {}
+    if pinned := _nonempty_model(manifest.get("default_model")):
+        logger.info("Resolved cycle model: %s (source=workflow:%s)", pinned, instance_config.workflow)
+        return pinned
+    logger.info("Resolved cycle model: %s (source=config.json)", global_config.model)
+    return global_config.model
```
petrsimon, line 293: Tier version of the resolver (see the design comment on `docs/presets/custom-workflows.md`). Only step 3 changes: the manifest names a tier, the map in `Config.model_tiers` turns it into an ID, and a tier that vanished from the map between cycles logs an error and falls through to the global default (startup validation catches the static case).

```suggestion
def resolve_cycle_model(
    script_dir: Path,
    instance_config: InstanceConfig,
    global_config: Config,
    remote_agent_dir: Path | None = None,
) -> str:
    """Resolve which model to use for the agent cycle following 4-tier precedence:

    1. instance.yaml `model` (explicit pin)
    2. BOT_MODEL environment variable (deploy overlay)
    3. workflow manifest.yaml `model_tier`, mapped through config.json `claude.modelTiers`
    4. global config.json `claude.model` (fallback)

    Presets name a tier, not a model ID, so shared workflows stay provider-neutral;
    only deployment-owned config (instance.yaml, BOT_MODEL, config.json) carries IDs.
    """
    logger = logging.getLogger(__name__)
    if pinned := _nonempty_model(instance_config.model):
        logger.info("Resolved cycle model: %s (source=instance.yaml)", pinned)
        return pinned
    if pinned := _nonempty_model(os.environ.get("BOT_MODEL")):
        logger.info("Resolved cycle model: %s (source=BOT_MODEL)", pinned)
        return pinned
    manifest = load_manifest(script_dir, instance_config.workflow, remote_agent_dir) or {}
    if tier := _nonempty_model(manifest.get("model_tier")):
        if pinned := global_config.model_tiers.get(tier):
            logger.info("Resolved cycle model: %s (source=workflow:%s tier=%s)", pinned, instance_config.workflow, tier)
            return pinned
        logger.error(
            "Workflow '%s' requests model tier '%s' not defined in config.json claude.modelTiers — using default",
            instance_config.workflow,
            tier,
        )
    logger.info("Resolved cycle model: %s (source=config.json)", global_config.model)
    return global_config.model
```

Three edits GitHub will not let me anchor here, all in the full patch in the review body:

- `Config` gets `model_tiers: dict[str, str] = field(default_factory=dict)` and `load_config` fills it from `claude.modelTiers` via a small `_load_model_tiers` helper (drops empty/non-string entries with a warning).
- `validate_manifest` takes `model_tiers: dict[str, str] | None = None` and adds `Model tier '<x>' not defined in config.json claude.modelTiers` to `errors` when the manifest names a tier that is not in the map. `None` keeps the old behavior for callers that do not pass it.
- `bot/run.py` passes `config.model_tiers` at the startup `validate_manifest` call.

docs/migrations/preset-migration-guide.md
petrsimon: This sentence contradicts the `BOT_MODEL` row directly above it.

```suggestion
`BOT_WORKFLOW_PRESET` and `BOT_ENV_PRESETS` are checked only when no `instance.yaml` is found; if `instance.yaml` exists, it takes precedence. `BOT_MODEL` also applies when `instance.yaml` exists but omits `model`.
```

(Adjust once the precedence question in `bot/config.py` is settled.)

docs/presets/custom-workflows.md
petrsimon: Design change request for this PR.

A concrete model ID here couples a shared workflow preset to one provider. Presets are otherwise provider-neutral (instructions, skills, preflights), while the model ID is a deployment concern: it has to match the proxy allowlist (`VERTEX_ALLOWED_MODELS`) and, under the OpenCode runtime (REHOR-137), the default becomes an OpenAI-family ID resolved by a different provider. A preset that says `claude-sonnet-4-6` would then be wrong or rejected there, and every manifest would need editing on a provider change.

Let the workflow declare *what it needs*, and keep the mapping in deployment-owned config next to the existing default:

```suggestion
# Optional model tier for cycles using this workflow. Tier names map to
# model IDs in config.json `claude.modelTiers`; presets never name a model ID.
model_tier: light
```

```json
"claude": {
  "model": "claude-opus-4-6",
  "modelTiers": { "light": "claude-sonnet-4-6", "heavy": "claude-opus-4-6" }
}
```

Resolution keeps your precedence: `instance.yaml model` (concrete pin, deploy knows its allowlist) > `BOT_MODEL` > `modelTiers[manifest.model_tier]` > `claude.model`. Unknown tier fails at startup, which also gives us the fail-fast this PR currently lacks. It lines up with the planned preflight complexity routing (REHOR-126), which would emit a tier rather than an ID, and with subagent tiers (REHOR-128).

The table row at line 112 and the "Model Resolution Order" section in `docs/presets/instance-config.md` need the same rename; both are in the full patch in the review body.

petrsimon (changes requested):
thanks a lot for picking it up, few ideas, please take a look

petrsimon (approved):
thanks a lot!

PR #520, opened 2026-09-15
feat(runtime): add TypeScript Claude Agent adapter

petrsimon:
Implement the TypeScript Claude Agent runtime adapter and wire it into the provider-neutral coordinator runtime registry.

Key changes:
- Add the `ClaudeAgentRuntime` adapter using `@anthropic-ai/claude-agent-sdk@0.3.270`.
- Normalize SDK lifecycle, model, tool, usage, runtime, and terminal events into Rehor events.
- Preserve session references, partial/final usage, estimated cost, cancellation, timeout, and cleanup behavior.
- Register the adapter as runtime ID `claude`.
- Forward Python-prepared MCP servers and allowed tools through the coordinator bridge.
- Add compatibility projection coverage for cost records and representative Claude cycles.
- Preserve partial usage and terminal events while draining timeout/cancellation cleanup.
- Aggregate usage per returned model, project turn-budget metrics, preserve result summaries, and match Python turn-budget hook behavior.
- Document runtime selection and configuration wiring.

Validation:
- `make verify` — passed.
- Python: 1071 passed, 3 skipped, 2 xfailed, 1 xpassed.
- Dashboard: 85 tests passed, typecheck/build passed.
- Coordinator: 86 tests passed, typecheck/build passed.
- `npm audit --audit-level=high` — 0 vulnerabilities.

This remains a migration boundary: the TypeScript adapter is production-ready behind the coordinator API, while the existing Python runner remains the current production entrypoint.

---

### Description

[REHOR-140](https://issues.redhat.com/browse/REHOR-140)

---

### Blast radius

Coordinator consumers only. Existing Python runtime behavior remains unchanged; the adapter is selected explicitly through runtime ID `claude`. Configuration forwarding is covered by coordinator bridge and runtime-factory tests.

---

### Rollback plan

Git revert is sufficient. The current production Python entrypoint is unchanged, and the TypeScript adapter is not enabled by the production runner.

---

### Checklist

### AI disclosure
Assisted by: [Pi](https://pi.dev) — gpt-5.6-luna xhigh


coordinator/src/ports/runtime-config.ts
```diff
@@ -0,0 +1,17 @@
+/** Provider-neutral MCP transport configuration forwarded to a runtime adapter. */
```
Hyperkid123, line 1: @petrsimon what do you think about using zod for validation? Or should the SDK do the validation on their own? From experience its kind of hard to deubug bad mcp configuration shape. If we could do it nicelly using zod before the agent session event starts and log it? We do not have to kill the entire process but its something alerting could be configured for via prometheus (eventually). And we would still get the types.

coordinator/src/runtimes/claude-agent.ts
Hyperkid123, line 1: Probably something for a future consolidation if we decide to keep claude (not sure of we do cause opencode can do it all), but could we re-use some of the sdk specific code between calude/opencode?

Because I think we will eventually throw it away, I am not going to spend much time on the claude binding.

Hyperkid123 (changes requested):
Findings:

- [P1] `coordinator/src/runtimes/claude-agent.ts:216-230`: SDK subprocess inherits `process.env` when `env` is omitted, exposing `GH_TOKEN`, `GITHUB_TOKEN`, signing keys, and SSO credentials to Claude tools. Mirror existing `sanitize_env()` behavior before enabling this runtime.
- [P2] `coordinator/src/cycle-input.ts:82-97`: `configHash` omits `mcpServers` and `allowedTools`; security/config changes can reuse stale run identity. Include both in hash.
- [P2] `coordinator/src/runtimes/claude-agent.ts:567-579`: Empty `modelUsage` prevents fallback to non-empty top-level `usage`, losing final token totals in some SDK results.

Verification:

- TypeScript tests: 87 passed.
- Typecheck and build passed.
- `git diff --check` passed.
- Python bridge tests: 4 passed; 2 blocked by missing `prometheus_client` dependency.

coordinator/src/runtimes/claude-agent.ts
petrsimon: Exactly my thoughts, i.e. we can extract only the shared lifecycle behavior later

coordinator/src/ports/runtime-config.ts
petrsimon: Yes, eventually for sure. We can do it with avj or switch to zod if we keep this runtime. Will gladly add it using ajv if you like, it shouldn't be too big.

Hyperkid123: Not this PR. I think runtime validation eventually would be good. Lets keep this focused.

petrsimon: 100% then :)

petrsimon: fyi doublecked if we are tracking it properly, some holes found, patched
 - REHOR-138 (https://issues.redhat.com/browse/REHOR-138) — runtime use of Rehor-owned JSON Schemas, provider-neutral config validation,
   redacted failures, and schema-drift fixtures.
 - REHOR-141 (https://issues.redhat.com/browse/REHOR-141) — parity fixtures for invalid MCP transports/tools, pre-session validation,
   coordinator survival, and no model startup.
 - REHOR-144 (https://issues.redhat.com/browse/REHOR-144) — already contains: “OpenCode configuration validation fails safely before a cycle
   starts.” Left unchanged.

petrsimon:
konflux working hard, hopefully it should be fine now 🤞

Hyperkid123 (approved):

PR #521, opened 2026-09-15
Add OpenCode V1 runtime adapter and supervisor

petrsimon:
## Summary

- Add the OpenCode V1 `AgentRuntime` adapter and runner-owned process supervisor for [REHOR-143](https://issues.redhat.com/browse/REHOR-143).
- Harden crash handling, bounded reconciliation, root/child session attribution, result filtering, completed-turn accounting, cumulative usage/cost snapshots, SDK error propagation, process-group cleanup, and output-pipe draining.
- Pin `@opencode-ai/sdk` to `1.18.29`.
- Keep packaging, OpenCode configuration/workflows, gateway integration, rollout, and production runner migration out of this change.

## Validation

- `make verify`
- Coordinator: 55 tests, typecheck, build
- Coordinator Biome check
- `npm audit --audit-level high`

---

### Blast radius

---

### Rollback plan

---

### Checklist
- [ ] No breaking changes to existing consumers (or migration path documented)
- [ ] No hardcoded secrets, tokens, or passwords

### AI disclosure

Assisted by: [Pi](https://pi.dev) — gpt-5.6-luna high


coordinator/src/runtimes/shared.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  } else if (name.includes("jira_transition_issue")) {
+    context.workType = context.workType ?? "new_ticket";
+  } else if (name.endsWith("memory_delete")) {
+    context.workType = context.workType ?? "memory_housekeeping";
+  }
+
+  const progress = asObject(input.progress);
+  if (progress) {
+    if (typeof progress.jira_key === "string" && progress.jira_key) {
```
Hyperkid123: This may be some artifact of the current SDK. We should not be directly referencing any of this. Workflows define the source of their work we should not assume a specific source. All entries should be already migrated. 

I don't think we need this fallback anymore. But we should probably have some followup cleanup task(s)

coordinator/src/runtimes/opencode-v1/runtime.ts
```diff
@@ -0,0 +1,1156 @@
...
+              responseStyle: "data",
+              throwOnError: true,
+            }),
+          active.controller.signal,
+          Date.now() + this.requestTimeoutMs,
+          "OpenCode session prompt",
+        ),
+        "OpenCode session prompt",
+      );
+      promptSubmitted = true;
```
Hyperkid123: **P1: Prompt submission can be accepted before this response resolves.** If `promptAsync` times out or the connection drops after OpenCode accepts the request, `promptSubmitted` remains false, so the catch path skips reconciliation and cleanup can delete the active session, losing result and usage data. Mark submission before issuing the request, or reconcile ambiguous prompt failures before deleting the session.

coordinator/src/runtimes/opencode-v1/process-supervisor.ts
```diff
@@ -0,0 +1,728 @@
...
+      throw new OpenCodeReadinessError(
+        `OpenCode config hash ${configHash ?? "missing"} does not match expected ${this.options.expectedConfigHash}`,
+      );
+    }
+
+    const allocatePortFn = this.options.allocatePort ?? allocatePort;
+    const port = this.options.port ?? (await allocatePortFn());
+    const environment = buildOpenCodeEnvironment(this.options);
+    if (this.options.config) {
+      environment.OPENCODE_CONFIG_CONTENT = stableJson(this.options.config);
```
Hyperkid123: **P1: Config integrity is optional while config content controls executable policy.** When `expectedConfigHash` is omitted, arbitrary `OPENCODE_CONFIG_CONTENT` can enable MCP servers, commands, filesystem access, or network behavior. Require a verified config hash or enforce a strict allowlist/schema before spawning OpenCode.

coordinator/src/runtimes/opencode-v1/environment.ts
```diff
@@ -0,0 +1,136 @@
...
+  "NO_COLOR",
+  "PATH",
+  "SHELL",
+  "TMPDIR",
+  "USER",
+  "XDG_CACHE_HOME",
+  "XDG_CONFIG_HOME",
+  "XDG_DATA_HOME",
+  "XDG_STATE_HOME",
+  "NODE_OPTIONS",
```
Hyperkid123: **P2: `NODE_OPTIONS` undermines this sanitized environment boundary.** Inherited values can preload arbitrary modules into the OpenCode/MCP child and alter runtime behavior before policy checks. Remove it from the default allowlist or validate it against an explicit policy.

coordinator/src/runtimes/opencode-v1/process-supervisor.ts
```diff
@@ -0,0 +1,728 @@
...
+              ),
+            ),
+          ),
+        Math.max(0, deadline - Date.now()),
+      );
+      child.stdout?.on("data", onData);
+      child.stderr?.on("data", onData);
+      child.once("exit", onExit);
+      child.once("error", onError);
+      signal.addEventListener("abort", onAbort, { once: true });
```
Hyperkid123: **P2: Startup abort race.** The abort listener is registered after port allocation and timer setup. If `signal` aborts before this line, `waitForListening()` misses the event and waits until startup timeout. Check `signal.aborted` immediately before registration and fail synchronously.

coordinator/src/runtimes/shared.ts
```diff
@@ -0,0 +1,208 @@
...
+      cleanup();
+      callback();
+    };
+    const onAbort = (): void => settle(() => reject(abortReason(operationSignal)));
+
+    if (operationSignal.aborted) {
+      onAbort();
+    } else {
+      operationSignal.addEventListener("abort", onAbort, { once: true });
+      pending = Promise.resolve().then(() => operation(operationSignal));
```
Hyperkid123: **P2: Cancellation does not prevent the queued operation from starting.** `operation()` is scheduled in a microtask after the abort listener is installed. An abort can reject this wrapper while the microtask still invokes the underlying session/create/prompt operation. Check `operationSignal.aborted` inside the microtask before calling `operation`.

coordinator/src/runtimes/opencode-v1/runtime.ts
```diff
@@ -0,0 +1,1156 @@
...
+            "OpenCode runtime ended without a completed result",
+        },
+        { runtimeSessionRef: active.sessionId },
+      );
+    }
+
+    if (failure && outcome.state !== "cancelled" && outcome.state !== "timed_out") {
+      yield factory(
+        "error",
+        { message: errorMessage(failure) ?? "OpenCode runtime failed" },
```
Hyperkid123: **P2: Sensitive provider/runtime output can be persisted without redaction.** `errorMessage(failure)` is emitted into the Rehor event stream; SDK/provider errors may include API keys, authorization headers, request bodies, or URLs with credentials. Add centralized secret redaction and tests covering sensitive values in provider and startup errors.

coordinator/src/runtimes/opencode-v1/process-supervisor.ts
```diff
@@ -0,0 +1,728 @@
...
+    const port = this.options.port ?? (await allocatePortFn());
+    const environment = buildOpenCodeEnvironment(this.options);
+    if (this.options.config) {
+      environment.OPENCODE_CONFIG_CONTENT = stableJson(this.options.config);
+    }
+    const args = ["serve", `--hostname=${this.options.hostname}`, `--port=${port}`];
+    const spawnProcess = this.options.spawnProcess ?? defaultSpawn;
+    this.startupOutput = "";
+    const child = spawnProcess(this.options.command, args, {
+      cwd: directory,
```
Hyperkid123: **P2: Worktree path is an unvalidated process boundary.** This caller-provided path becomes child `cwd` and is also sent as the OpenCode directory. Validate absolute path, resolved realpath, approved workspace root, and ownership before spawning the child to prevent execution against unintended host files.

Hyperkid123 (commented):
Stuff

Hyperkid123 (commented):
some followups we should track

Hyperkid123 (approved):

PR #529, opened 2026-09-18
feat(opencode): deterministic OpenCode V1 configuration

petrsimon:
## Summary

Implements the deterministic OpenCode V1 configuration/workflow boundary for [REHOR-144](https://issues.redhat.com/browse/REHOR-144).

- Renders and validates provider/model, MCP, permissions, secret references, plugins, and exact package closure.
- Preserves generated `CLAUDE.md`, persona instruction assembly, and project `.claude/skills` discovery.
- Carries reference-only MCP configuration across the Python/coordinator bridge, with explicit optional MCP metadata and strict required-server validation.
- Isolates per-cycle OpenCode config/state, disables runtime downloads/autoupdate, and validates before supervisor startup.
- Aligns prompt submission and event attribution with the effective rendered model.
- Keeps Claude as the default production runtime; OpenCode registration remains explicit and opt-in.

## Validation

- `make verify` passed.
- Coordinator: 160 tests, typecheck, and build passed.
- Python: 1,072 passed, 3 skipped, 2 xfailed, 1 xpassed.
- Dashboard: typecheck, build, and 85 tests passed.
- Go vet and race tests passed.
- Exact OpenCode `1.18.29` isolated health and project-skill discovery smokes passed.

Provider descriptor integration remains coordinated with [REHOR-145](https://issues.redhat.com/browse/REHOR-145); package pre-bake remains owned by [REHOR-142](https://issues.redhat.com/browse/REHOR-142); production canary enablement remains out of scope for [REHOR-146](https://issues.redhat.com/browse/REHOR-146).

---

### Description

Adds a validated, deterministic OpenCode V1 configuration renderer and runtime handoff while leaving the existing Claude/Vertex production path unchanged.

[REHOR-144](https://issues.redhat.com/browse/REHOR-144)

---

### Blast radius

Coordinator and Python bridge code gain an opt-in OpenCode configuration path. Existing Claude production selection and lifecycle remain unchanged. Production packaging and canary configuration are not changed.

---

### Rollback plan

The OpenCode factory is opt-in, so production behavior can be restored by leaving it unregistered or reverting this PR. No data migration or persistent-state rollback is required.

---

### Checklist
- [ ] Tested against at least one consuming repo/service (OpenCode remains opt-in; exact artifact smoke completed)
- [x] Container images pinned to specific tags, not `latest` (no image changes)

### AI disclosure
Assisted by: [Pi](https://pi.dev) — gpt-5.6-luna max


petrsimon:
@Hyperkid123 take it apart :)  Preparing rehor-146 stacked on top of this and going to test it with my keys 🤞

coordinator/src/runtimes/opencode-v1/config.ts
```diff
@@ -0,0 +1,1002 @@
...
+    }
+    permissions[mapped] = "allow";
+  }
+
+  for (const tool of BUILTIN_TOOLS) {
+    if (permissions[tool] === undefined) permissions[tool] = "deny";
+  }
+  for (const server of [...configuredMcpServers].sort()) {
+    if (!allowedMcpWildcards.has(server) && permissions[`${server}_*`] === undefined) {
+      permissions[`${server}_*`] = "deny";
```
Hyperkid123: This ordering looks like it will deny the MCP tools we explicitly allowed above. OpenCode resolves matching permission patterns last-match-wins, so for `{"jira_search": "allow", "jira_*": "deny"}` the wildcard deny wins. Could we emit the server wildcard denies before the specific MCP grants?

coordinator/src/runtimes/opencode-v1/config.ts
```diff
@@ -0,0 +1,1002 @@
...
+    );
+  }
+  return values;
+}
+
+function isCoordinatorControlledEnvironment(name: string): boolean {
+  return OPENCODE_BLOCKED_PASSTHROUGH_PREFIXES.some((prefix) => name.startsWith(prefix));
+}
+
+function assertProviderEnvironment(name: string, path: string): void {
```
Hyperkid123: Could we make this an allowlist rather than only checking known blocked names? Any provider/plugin `{env:NAME}` reference becomes a passthrough variable, so config can currently request arbitrary values such as `AWS_SECRET_ACCESS_KEY` or `DATABASE_URL`. Since this is an environment boundary, I would rather fail closed for anything outside an explicit safe set.

Hyperkid123 (changes requested):
I went through this locally and ran the coordinator tests/typecheck/build plus the Python suite. I found two things worth fixing before merging:

Hyperkid123 (approved):

